PR #5, opened 2022-10-05
Fix some clang warnings

trenouf:


lib/TableGen/GenDialect.cpp
```diff
@@ -841,15 +843,15 @@ void DialectsContext::init(RecordKeeper& records,
       return true;
     };
 
-    for (const auto result : llvm::enumerate(op->results)) {
+    for (const auto &result : llvm::enumerate(op->results)) {
```
nhaehnle, line 846: I thought llvm::enumerate doesn't produce lvalues. But I suppose a const-reference can bind to a temporary and extend its lifetime. It's interesting that Clang is opinionated about that...

nhaehnle (approved):
LGTM

PR #6, opened 2022-10-05
Export llvm-dialects's include path

trenouf:
... for client projects to be able to access it without having to explicitly add it to their own target_include_directories.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,8 @@ else()
     target_link_libraries(llvm-dialects-tblgen PRIVATE llvm_dialects_tablegen ${llvm_libs})
 endif()
 
+target_include_directories(llvm_dialects PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+
 target_sources(llvm_dialects PRIVATE
     lib/Dialect/Builder.cpp
     lib/Dialect/Dialect.cpp


CMakeLists.txt
```diff
@@ -63,6 +63,8 @@ else()
     target_link_libraries(llvm-dialects-tblgen PRIVATE llvm_dialects_tablegen ${llvm_libs})
 endif()
 
+target_include_directories(llvm_dialects PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
```
Flakebi, line 66: I guess the `include_directories(include)` above can be removed then?
Also, `${CMAKE_CURRENT_SOURCE_DIR}` could be removed, it’s the default for relative paths.

trenouf: The `include_directories(include)` is needed by the other two targets in this file. Should we instead give each of the three targets a `target_include_directories(.. PRIVATE` ?

Flakebi: Yes, that would be nice.
`include_directories` is not deprecated, but the `target_*` functions are preferred in the CMake docs.

nhaehnle:
It should be enabled now.

trenouf:
> It should be enabled now.

Thanks; that worked.

PR #7, opened 2022-10-06
Handle op with no args

trenouf:
Dunno if it's just my compiler (clang 13), but it refused to implicitly construct ArrayRef<Value*> out of an empty array of args.


nhaehnle (approved):
Usually it's possible to initialize an empty `ArrayRef` with `{}`, but in this case it looks like this might cause ambiguous overloads in the IRBuilder::CreateCall overload set. LGTM

PR #14, opened 2023-01-11
Add dependency libs

Flakebi:
Add dependencies on LLVM libs.
Fixes building with shared libraries on.

Also move some common code out of the if.


nhaehnle:
Is that actually common code? I remember there are some issues with using `add_llvm_library` out-of-tree.

Flakebi:
Oh, I misread the `add_library` for `add_llvm_library`…

nhaehnle (approved):
LGTM

PR #15, opened 2023-01-16
Use std::optional instead of llvm::Optional

jayfoad:


jayfoad:
This is required to fix some of our internal builds which broke when upstream LLVM stopped including llvm/ADT/Optional.h from some other headers - see llvm/llvm-project@da6642a1c81bb5ff68153bcdc138e75e6ecdce43

nhaehnle (approved):
LGTM

PR #17, opened 2023-01-17
Improve some comments in GenDialect.cpp

tsymalla:
I corrected some error messages which could be misleading.

### DIFF
--- a/lib/TableGen/GenDialect.cpp
+++ b/lib/TableGen/GenDialect.cpp
@@ -884,7 +884,7 @@ std::unique_ptr<PredicateExpr> DialectsContext::parsePredicateExpr(DagInit *dag)
 
       if (!isa<DagInit>(arg)) {
         report_fatal_error(
-            Twine("Logical expression has missing or non-dag argument: ")
+            Twine("Logical expression is missing argument or has non-dag argument: ")
                 + dag->getAsString());
       }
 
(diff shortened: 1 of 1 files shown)


lib/TableGen/GenDialect.cpp
```diff
@@ -571,7 +571,9 @@ class SymbolTable {
   std::unordered_set<std::string> m_names;
 
 public:
-  std::string chooseName(StringRef name) { return chooseName(ArrayRef(name)); }
+  std::string chooseName(StringRef name) {
+    return chooseName(makeArrayRef(name));
+  }
```
Flakebi, line 576: Why did you revert this change?

tsymalla: I did not revert this intentionally. I just used the GitHub web editor. Maybe the fork being generated is not up-to-date.

Flakebi: Sounds like a race-condition in GitHub’s web ui. (I.e. start editing from a certain state, but once one presses save, GitHub gets the new state, which is different when someone pushed in-between.)

tsymalla: Yes. Sounds likely, and I rebased onto recent dev.

nhaehnle (approved):
Thanks!

tsymalla:
@nhaehnle You're welcome. I guess it is currently you taking care of merging PRs?

PR #18, opened 2023-01-26
Support memory(...) attributes, and plenty of refactoring

nhaehnle:
The main functional change of this series is to support memory(...) attributes by writing a trait like `Memory<[]>` instead of ReadNone or `Memory<[(ref ArgMem)]>` instead of a ReadArgMemOnly.

The series also refactors GenDialect.cpp into smaller files to get a clearer split into "model" and "view/printer", adds some AttributeList caching for compile time and simplifies the function types we use via varargs (idea shamelessly stolen from BuilderRecorder).


lib/TableGen/Dialects.cpp
```diff
@@ -26,6 +26,59 @@
...
+        return 1;
+
+      StringRef lhsName = (*lhsIt)->getName();
+      StringRef rhsName = (*rhsIt)->getName();
+      int compare = lhsName.compare(rhsName);
+      if (compare != 0)
+        return compare;
+      ++lhsIt;
+      ++rhsIt;
+    }
```
jasilvanus, line 53: I think this can be replaced by `std::lexicographical_compare`.

jasilvanus (commented):
I read all of it, but lack a bit of context assess the patch as a whole.

A high-level question only slightly related to this particular patch: Shouldn't we have lit tests for the TableGen parts to check the generated C++ sources, similar to  `llvm/test/TableGen` in `llvm-project`? I find the source generation a bit obfuscated in some parts.

include/llvm-dialects/Dialect/Dialect.td
```diff
@@ -197,15 +197,40 @@ def AttrI64 : IntegerAttr<"uint64_t", I64>;
...
+/// Example: `Memory<[(ref), (mod ArgMem, InaccessibleMem)]>` means the
+/// operation may read from any kind of memory and write to argument and
+/// inaccessible memory.
+class Memory<list<dag> effects_> : Trait {
+  list<dag> effects = effects_;
+}
+
+def ref;
+def mod;
+def modref;
```
Flakebi, line 225: Maybe read, write and readwrite is easier to understand? (I now found these defined in [MemoryEffects](https://llvm.org/doxygen/namespacellvm.html#a9db8c44b250b90e3ab7e4d144e7c9c2e); as you said offline, the textual IR calls them read and write)

Flakebi (approved):
LGTM, one comment inline

nhaehnle:
> A high-level question only slightly related to this particular patch: Shouldn't we have lit tests for the TableGen parts to check the generated C++ sources, similar to `llvm/test/TableGen` in `llvm-project`? I find the source generation a bit obfuscated in some parts.

I'm not a fan of the idea of matching the generated C++ with lit tests. Maybe there are particular corner cases where it makes sense, but what I would generally prefer is:

1. Write a small test/example program that defines random dialects and exercises them, add it to CI here.
2. Prettify the generated code somehow. (Perhaps it makes sense to add some functionality for that to upstream llvm libLLVMTableGen and then use it from here?)
3. Figure out if we can get CI to produce a diff of the generated C++ of the test/example program as a reviewing aid.

jasilvanus:
My intention was being able to see the impact of implementation changes on the generated C++ source. So yes, it does not have to be lit tests (maybe explain why not?), but I think there should be some diff of generated C++ visible for review.

Also, for NFC commits it would increase the confidence level that indeed the generated code matches.

nhaehnle:
> why not?

Good lit tests for generated C++ code are just hard to write and keep up-to-date. LLVM has been moving to update_test_checks which basically just checks in the tool output. So maybe that's the solution: Set up a flow where the generated C++ for the "test"/"example" use is just checked into the Git repository, and part of the test/example build setup diffs the generated C++ from the build directory against the checked-in version. Then updating it becomes trivial, and you can easily see changes as part of the PR. It doesn't use lit, but at that point I don't think using lit has any benefit?

PR #19, opened 2023-01-30
Fix warnings about lack of ArrayRef deduction guides

jayfoad:
Fix warnings when built against an older LLVM that does not have deduction guides for ArrayRef:

warning: 'ArrayRef' may not intend to support class template argument deduction [-Wctad-maybe-unsupported]


nhaehnle (approved):
Huh okay, seems we were to eager here :)

PR #23, opened 2023-02-07
Add -no-memory-effects option

nhaehnle:
This is a temporary workaround for compatibility with not-quite-recent versions of LLVM.

### DIFF
--- a/test/example/CMakeLists.txt
+++ b/test/example/CMakeLists.txt
@@ -1,7 +1,7 @@
 
 ### Example application
 
-add_executable(llvm-dialects-example
+add_executable(llvm-dialects-example EXCLUDE_FROM_ALL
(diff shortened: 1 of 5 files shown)


test/example/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 
 ### Example application
 
-add_executable(llvm-dialects-example
+add_executable(llvm-dialects-example EXCLUDE_FROM_ALL
```
Flakebi, line 4: You can also put the `EXCLUDE_FROM_ALL` onto the [`add_subdirectory`](https://cmake.org/cmake/help/latest/command/add_subdirectory.html) for the example folder, then you don’t need to add it to every target created here.

PR #26, opened 2023-02-22
Add a verifier and verifier pass

nhaehnle:
This PR is split into a bunch of commits which first add some generic "extension" facilities and then build the verifier on top of it. The verifier pass currently checks precisely the constraints on operations that are also checked by the builder methods.

### DIFF
--- a/example/ExampleMain.cpp
+++ b/example/ExampleMain.cpp
@@ -26,13 +26,32 @@
 #include "ExampleDialect.h"
 
 #include "llvm-dialects/Dialect/Builder.h"
+#include "llvm-dialects/Dialect/Verifier.h"
 
-#include "llvm/IR/Module.h"
+#include "llvm/AsmParser/Parser.h"
 #include "llvm/IR/DerivedTypes.h"
+#include "llvm/IR/Module.h"
 #include "llvm/IRPrinter/IRPrintingPasses.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/SourceMgr.h"
 
 using namespace llvm;
 using namespace llvm_dialects;
+
+enum class Action {
+  Build,
+  Verify,
+};
+
+cl::opt<Action> g_action(
+    cl::desc("Action to perform:"), cl::init(Action::Build),
+    cl::values(clEnumValN(Action::Build, "build", "Example IRBuilder use"),
+               clEnumValN(Action::Verify, "verify", "Verify an input module")));
+
+// Input sources
+cl::list<std::string> g_inputs(cl::Positional, cl::ZeroOrMore,
+                               cl::desc("Input file(s) (\"-\" for stdin)"));
 
 void createFunctionExample(Module &module, const Twine &name) {
   Builder b{module.getContext()};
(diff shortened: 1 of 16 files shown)


include/llvm-dialects/Dialect/ContextExtension.h
```diff
@@ -0,0 +1,112 @@
...
+public:
+  class Key : public detail::ContextExtensionKey {
+    static void *createImpl(llvm::LLVMContext &context) {
+      return static_cast<void *>(new ContextExtensionT(context));
+    }
+
+    static void destroyImpl(llvm::LLVMContext &context, void *untyped) {
+      ContextExtensionT *extension = static_cast<ContextExtensionT *>(untyped);
+      delete extension;
+    }
```
jasilvanus, line 91: Is the manual handling of function pointers really necessary? 

What if we introduced a common, non-templated base class `ContextExtensionImplBase` of all ContextExtensionImpl that has a virtual destructor,  and replaced the `void*` by  `std::unique_ptr<ContextExtensionImplBase>`? 

For creation, we could use a virtual function in `Key`.

example/ExampleMain.cpp
```diff
@@ -72,12 +91,40 @@ std::unique_ptr<Module> createModuleExample(LLVMContext &context) {
...
+    if (!module) {
+      error.print("llvm-dialects-example", errs());
+      errs() << "\n";
+      return 1;
+    }
+
+    if (!verify(*module, outs()))
+      return 1;
+  } else {
+    llvm_unreachable("unhandled action");
```
jasilvanus, line 126: Maybe use `report_fatal_error` instead? Per my understand, `llvm_unreachable` does not guarantee error messages for the benefit of better performance, but that should not matter here.

include/llvm-dialects/Dialect/ContextExtension.h
jasilvanus: We could also use a virtual function for `destroy` even without a virtual destructor, but also getting rid of the manual memory management and `void*` casts may be even nicer?

lib/Dialect/ContextExtension.cpp
```diff
@@ -0,0 +1,78 @@
...
+ContextExtensionKey::ContextExtensionKey(CreateFn *createFn,
+                                         DestroyFn *destroyFn)
+    : m_create(createFn), m_destroy(destroyFn) {
+  auto &keys = getContextExtensionKeys();
+
+  for (auto enumeratedKey : llvm::enumerate(keys)) {
+    if (!enumeratedKey.value()) {
+      enumeratedKey.value() = this;
+      m_index = enumeratedKey.index();
+      goto inserted;
```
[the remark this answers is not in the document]
nhaehnle: Returning from a constructor feels weird. Anyway, I'll rewrite it.

example/ExampleMain.cpp
nhaehnle: This is more like an assertion because the enum of actions should be closed. But it doesn't really matter.

lib/Dialect/ContextExtension.cpp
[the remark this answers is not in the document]
jasilvanus: You could also do something like

```
m_index = [&](){
   for (auto enumeratedKey : llvm::enumerate(keys) {
      if (!enumeratedKey.value()) {
         enumeratedKey.value() = this;
         return enumeratedKey.index();
      }
      keys.push_back(this);
      return keys.size();
   }
}();

include/llvm-dialects/Dialect/ContextExtension.h
nhaehnle: Yeah, turns out that this can work out nicely after all.

jasilvanus: Nice!

jasilvanus (approved):

PR #32, opened 2023-02-24
Add CI

Flakebi:
Add a CI based on the docker images built for amdvlk.
I wanted to use apt.llvm.org first, but we don’t have any control over
the llvm commit used, which would regularly break the CI build when
llvm’s C++ API changes.

This CI build fetches the amdvlk base docker image, updates the
llvm-dialects submodule and runs the tests.

Also, fix compiling with shared libraries. A reference to AsmParser was
missing and add an assert that checks for out-of-bounds vector accesses.
The vector is sometimes accessed out-of-bounds, leading to inconsistent
behavior (like asserting in debug mode, but running through when run in a debugger).

The build here will not pass, because of the afore mentioned oob access, here’s a passing build with a workaround: https://github.com/Flakebi/llvm-dialects/actions/runs/4264669825


jasilvanus:
I welcome the change, but am not an expert on the CI details by any means, so let's rely on what @nhaehnle has to say.

PR #33, opened 2023-02-24
Allow unnamed types

Flakebi:
Allow intrinsics with unnamed struct types.
The problem with unnamed struct types, is that mangling based on the name does not work well, because multiple types map to the same mangled name and getting a function with a mangled name then returns a constantexpr bitcast of the existing function with the wrong type.

I see multiple ways to solve this:

1. After getting a function, check if it has the correct type and if it has the wrong type, append a suffix. Repeat until a function with the correct type is found or created. This is the easiest approach and implemented here.
2. Use a HashMap to map each function type to a name and use the map for fast lookups and to find unique suffixes. This shines when the first approach would have a lot of name collisions, but it needs memory and is the most complex approach to implement.
3. Same as 1., use linear probing in case of a name collision, but to reduce the chance of a conflict, choose the initial name with a random 2-digit suffix.


lib/TableGen/Operations.cpp
```diff
@@ -459,6 +459,11 @@ void BuilderMethod::emitDefinition(raw_ostream &out, FmtContext &fmt,
   fmt.addSubst("fn", symbols.chooseName({"fn", "function"}));
   out << tgfmt(R"(
     auto $fn = $_module.getOrInsertFunction($fnName, $fnType, $attrs);
+    for (unsigned i = 0; ::llvm::isa<::llvm::ConstantExpr>($fn.getCallee()); i++) {
```
jasilvanus, line 462: Can't we instead check for `!::llvm::isa<::llvm::Function>`?
Either way, please add a comment on why we are checking the type of the `getCallee()`.

lib/TableGen/Operations.cpp
```diff
@@ -459,6 +459,11 @@ void BuilderMethod::emitDefinition(raw_ostream &out, FmtContext &fmt,
...
+      // Try new names until we get one with the correct type
+      std::string newName = (::llvm::Twine($fnName) + "_" + ::llvm::utostr(i)).str();
```
jasilvanus, line 464: Not sure whether it's worth it, but we could use a `SmallString` with a `llvm::raw_svector_ostream` instead.

nhaehnle:
Yeah, this is annoying but I think this approach is fine.

nhaehnle:
Can you please add a test for to the builder test? The existing ReadOp should be suitable.

Flakebi:
Thanks for the reviews.
I turned out that with opaque pointers, `Module::getOrInsertFunction()` never returns a bitcast, because it compares pointer types. Instead, it returns a function with the wrong type.

lib/TableGen/Operations.cpp
```diff
@@ -459,6 +459,20 @@ void BuilderMethod::emitDefinition(raw_ostream &out, FmtContext &fmt,
   fmt.addSubst("fn", symbols.chooseName({"fn", "function"}));
   out << tgfmt(R"(
     auto $fn = $_module.getOrInsertFunction($fnName, $fnType, $attrs);
+    ::llvm::SmallString<32> newName;
+    for (unsigned i = 0; ::llvm::cast<::llvm::Function>($fn.getCallee())->getFunctionType() != $fn.getFunctionType(); i++) {
+      // If a function with the same name but a different types already exists,
+      // we get a bitcast of a function.
```
jasilvanus, line 465: The comment regarding the bitcast seems to be outdated.

test/example/test-builder.test
```diff
@@ -23,5 +23,11 @@
 ; CHECK-NEXT:    [[TMP14:%.*]] = call ptr @xd.read.p0()
 ; CHECK-NEXT:    [[TMP15:%.*]] = call i8 (...) @xd.stream.add.i8(ptr [[TMP14]], i64 14, i8 0)
 ; CHECK-NEXT:    call void (...) @xd.write(i8 [[TMP15]])
+; CHECK-NEXT:    [[TMP16:%.*]] = call [[TMP0]] @xd.read.s_s()
+; CHECK-NEXT:    [[TMP17:%.*]] = call [[TMP1]] @xd.read.s_s_0()
+; CHECK-NEXT:    [[TMP18:%.*]] = call [[TMP2]] @xd.read.s_s_1()
```
nhaehnle, line 28: Hilarious UTC matches there....

lib/TableGen/Operations.cpp
```diff
@@ -459,6 +459,21 @@ void BuilderMethod::emitDefinition(raw_ostream &out, FmtContext &fmt,
...
+      // If a function with the same name but a different types already exists,
+      // we get a bitcast of a function or a function with the wrong type.
+      // Try new names until we get one with the correct type.
+      newName = "";
+      ::llvm::raw_svector_ostream newNameStream(newName);
+      newNameStream << $fnName << "_" << i;
+      fn = $_module.getOrInsertFunction(newNameStream.str(), $fnType, $attrs);
+    }
+    assert(::llvm::isa<::llvm::Function>($fn.getCallee()));
+    assert(!::llvm::isa<::llvm::ConstantExpr>($fn.getCallee()));
```
nhaehnle, line 474: The second assertion doesn't really make sense anymore, does it?

Flakebi:
I also added a test with typed pointers, which triggers a
different path when creating a function (with typed pointers, a bitcast
is returned; with opaque pointers, a function with a wrong type is
returned).

jasilvanus (approved):
LGTM

nhaehnle:
What about that second assertion?

CMakeLists.txt
```diff
@@ -10,6 +10,15 @@ project(llvm-dialects)
 
 message(STATUS "Begin configuring llvm-dialects...")
 
+function(set_compiler_options PROJECT_NAME)
```
tsymalla, line 13: Maybe name this enable_compiler_color_output? This is not really setting anything more than that

nhaehnle (commented):
Update the generated files? Thankfully there's a PR with CI enablement ;) Apart from that it LGTM.

nhaehnle (approved):

PR #35, opened 2023-02-28
Refactor type to string in Utils.

tsymalla:
Extracted the switch case into a separate function to cleanup `getMangledTypeStr` a bit.


nhaehnle (changes requested):
Please no, because this function is copied essentially verbatim from llvm-project as the comment indicates. Let's leave it as is so it easier to keep in sync.

nhaehnle:
Of course, if you feel like doing the same cleanup upstream and then porting it here, that would be fine.

tsymalla:
Okay! Thanks. :)

PR #37, opened 2023-03-03
Work around Clang warning about CTAD

nhaehnle:
It is almost certainly the intention that std::lock_guard supports CTAD, but with -Wctad-maybe-unsupported (which LLVM enabled as a conscious decision), Clang warns about it.

Fixes: https://github.com/GPUOpen-Drivers/llvm-dialects/issues/36


trenouf (approved):
LGTM thanks.

PR #39, opened 2023-03-17
More visitor features

nhaehnle:
The overall goal here is to have a unified framework for visiting instructions, whether they're dialect ops or core instructions, plus then a bunch of convenience features.

* Visitor: Add support for nested visitor clients
* Visitor: Add VisitorStrategy::ReversePostOrder
* Visitor: allow visiting core instructions and intrinsics
* Visitor: Optimize core opcode and intrinsic ID lookup
* Visitor: Add convenient way of using member functions

### DIFF
--- a/example/ExampleMain.cpp
+++ b/example/ExampleMain.cpp
@@ -146,6 +146,10 @@ struct VisitorInnermost {
 struct VisitorNest {
   raw_ostream *out = nullptr;
   VisitorInnermost inner;
+
+  void visitBinaryOperator(BinaryOperator &inst) {
+    *out << "visiting BinaryOperator: " << inst << '\n';
+  }
 };
 
 struct VisitorContainer {
@@ -179,9 +183,7 @@ template <bool rpot> const Visitor<VisitorContainer> &getExampleVisitor() {
                 [](VisitorNest &self, UnaryInstruction &inst) {
                   *self.out << "visiting UnaryInstruction: " << inst << '\n';
                 });
-            b.add<BinaryOperator>([](VisitorNest &self, BinaryOperator &inst) {
-              *self.out << "visiting BinaryOperator: " << inst << '\n';
-            });
+            b.add(&VisitorNest::visitBinaryOperator);
             b.nest<raw_ostream>([](VisitorBuilder<raw_ostream> &b) {
               b.add<xd::WriteOp>([](raw_ostream &out, xd::WriteOp &op) {
                 out << "visiting WriteOp: " << op << '\n';
--- a/example/ExampleMain.cpp
+++ b/example/ExampleMain.cpp
@@ -30,6 +30,8 @@
 
 #include "llvm/AsmParser/Parser.h"
 #include "llvm/IR/DerivedTypes.h"
+#include "llvm/IR/IntrinsicInst.h"
+#include "llvm/IR/Intrinsics.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IRPrinter/IRPrintingPasses.h"
 #include "llvm/Support/CommandLine.h"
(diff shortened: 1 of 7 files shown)


example/ExampleMain.cpp
```diff
@@ -172,9 +176,7 @@ void exampleVisit(Module &module) {
         b.add<UnaryInstruction>([](VisitorNest &self, UnaryInstruction &inst) {
           *self.out << "visiting UnaryInstruction: " << inst << '\n';
         });
-        b.add<BinaryOperator>([](VisitorNest &self, BinaryOperator &inst) {
-          *self.out << "visiting BinaryOperator: " << inst << '\n';
-        });
+        b.add(&VisitorNest::visitBinaryOperator);
```
Flakebi, line 179: To encourage the strongly encouraged pattern where a visitor is constructed once as a static variable, maybe the example should show how that works?

Flakebi (approved):
Didn’t look into all the details, but it looks good from what I saw.

example/ExampleMain.cpp
```diff
@@ -132,6 +139,75 @@ std::unique_ptr<Module> createModuleExampleTypedPtrs(LLVMContext &context) {
...
+  builder.nest<VisitorNest>(
+      [](VisitorBuilder<VisitorNest> &b) {
+        b.add<xd::ReadOp>([](VisitorNest &self, xd::ReadOp &op) {
+          *self.out << "visiting ReadOp: " << op << '\n';
+        });
+        b.add<UnaryInstruction>([](VisitorNest &self, UnaryInstruction &inst) {
+          *self.out << "visiting UnaryInstruction: " << inst << '\n';
+        });
+        b.add(&VisitorNest::visitBinaryOperator);
+        b.nest<raw_ostream>([](VisitorBuilder<raw_ostream> &b) {
```
jasilvanus, line 180: Maybe rename the builders to overload `b`, e.g. `nestedBuilder` and `innerStreamBuilder`?

jasilvanus (commented):
I skipped some details of the nested builder contruction. Some inline nits, but the general approach looks good to me.

example/ExampleMain.cpp
nhaehnle: There's a tension between doing that and the fact that this visitor build process is actually configurable by an argument.

example/ExampleMain.cpp
nhaehnle: Hmm, I kind of like it that way :)

include/llvm-dialects/Dialect/Visitor.h
```diff
@@ -45,37 +53,211 @@ enum class VisitorStrategy {
...
 
 private:
-  VisitorStrategy m_strategy = VisitorStrategy::ByFunctionDeclaration;
-  llvm::SmallVector<VisitorCase> m_cases;
+  VisitorBuilderBase(const VisitorBuilderBase &rhs) = delete;
+  VisitorBuilderBase &operator=(const VisitorBuilderBase &rhs) = delete;
+
+  VisitorTemplate m_ownedTemplate;
+  VisitorTemplate *m_template;
+  ssize_t m_projection;
```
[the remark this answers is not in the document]
nhaehnle: I'm encapsulating it in a wrapper.

example/ExampleMain.cpp
Flakebi: If `setStrategy` returns the builder again, shouldn’t it work like this?
```c++
static VisitorBuilder<VisitorContainer> builder = VisitorBuilder<VisitorContainer>()
  .nest<VisitorNest>(…)
  .setStrategy(g_rpot ? VisitorStrategy::ReversePostOrder : VisitorStrategy::Default)
  .build();
```

jasilvanus: This might work in this example, but in general I'm not sure initializing static objects with mutable global state is something we'd recommend, given the possibility of that state changing and then having a stale static object?

You could of course use *different* builders for the different strategies which would avoid that (wrapped e.g. in a templated getter), but I'm not sure that's worth the effort?

nhaehnle: Ah sure, why not :)

PR #41, opened 2023-04-06
WIP: tablegen: add hasTypeInfo field to DialectType

nhaehnle:
TODO:
- align with the required upstream changes ([Discourse](https://discourse.llvm.org/t/rfc-target-type-classes-for-extensibility-of-llvm-ir/69813), [Phabricator](https://reviews.llvm.org/D147697))

This allows users to hook into the (proposed) TargetExtTypeClass infrastructure to set the properties of dialect types.

Note: This is a single commit on top of #40, since I wanted to avoid the hassle of dealing with textual conflicts in the example and its tests.

### DIFF
--- a/test/example/generated/ExampleDialect.cpp.inc
+++ b/test/example/generated/ExampleDialect.cpp.inc
@@ -12,6 +12,8 @@
 
 #include "llvm/Support/ModRef.h"
 
+#include "llvm/IR/TargetExtType.h"
+    
 #include "llvm/Support/raw_ostream.h"
 #endif // GET_INCLUDES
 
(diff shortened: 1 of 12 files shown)


test/example/generated/ExampleDialect.cpp.inc
```diff
@@ -138,7 +140,14 @@ attrBuilder.addAttribute(::llvm::Attribute::NoUnwind);
...
 m_attributeLists[3] = ::llvm::AttributeList::get(context, ::llvm::AttributeList::FunctionIndex, attrBuilder);
 }
-}
+
+    static const ::llvm::TargetExtTypeClass classXdVectorType(
+          "xd.vector", false,
+          [](const ::llvm::TargetExtType *type) {
+            return ::llvm::cast<XdVectorType>(type)->getTypeInfo();
+          });
+    context.registerTargetExtTypeClass(&classXdVectorType);
```
jasilvanus, line 149: Shouldn't we only register upon the first pass?

Currently we re-create a dialect every time a DialectContext is created for it (not sure whether that's necessary), so we'd register the same class multiple times.

nhaehnle: The registration is per-LLVMContext, and we should really only be creating a DialectContext once per LLVMContext.

I suppose that this wasn't a strict rule previously, and now it becomes one, but I think it's a reasonable rule to have.

PR #42, opened 2023-05-12
Add .gitignore file.

tsymalla:
The file contents are mostly copied from upstream LLVM modulo some files and folders that are not relevant for llvm-dialects.


nhaehnle (approved):
Sure, makes sense to me.

PR #43, opened 2023-05-12
Extract operand checks in ConstraintSystem::addConstraintImpl

tsymalla:
Removes a bit of code duplication.

### DIFF
--- a/lib/TableGen/Constraints.cpp
+++ b/lib/TableGen/Constraints.cpp
@@ -84,19 +84,28 @@ bool ConstraintSystem::addConstraintImpl(raw_ostream &errs, Init *init,
                                          Variable *self) {
   if (auto *dag = dyn_cast<DagInit>(init)) {
     Record *op = dag->getOperatorAsDef({});
+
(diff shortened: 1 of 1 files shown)


nhaehnle (approved):
LGTM

lib/TableGen/Constraints.cpp
```diff
@@ -84,19 +84,28 @@ bool ConstraintSystem::addConstraintImpl(raw_ostream &errs, Init *init,
                                          Variable *self) {
   if (auto *dag = dyn_cast<DagInit>(init)) {
     Record *op = dag->getOperatorAsDef({});
+
+    auto isValidOperand = [&dag, &op, &errs](size_t index,
```
trenouf, line 88: Warning when building with clang:

```
llvm-dialects/lib/TableGen/Constraints.cpp:88:35: warning: lambda capture 'op' is not used [-Wunused-lambda-capture]
    auto isValidOperand = [&dag, &op, &errs](size_t index,
                               ~~~^~
```

PR #45, opened 2023-05-12
Enumerate template specializations in OpDescription via macro

tsymalla:
Remove the `template <> const OpDescription &OpDescription::get<...>()` specializations and use some macros instead.

I abstained from adding more complicated macros for now to handle the `Intrinsic::` prefix similarly for the non-array and array cases. Open for suggestions.

### DIFF
--- a/lib/Dialect/OpDescription.cpp
+++ b/lib/Dialect/OpDescription.cpp
@@ -123,208 +123,98 @@ template <> const OpDescription &OpDescription::get<BinaryOperator>() {
   }
 #include "llvm/IR/Instruction.def"
 
+#define HANDLE_INTRINSIC_DESC(Class, opcode)                                   \
+  template <> const OpDescription &OpDescription::get<Class>() {               \
+    static const OpDescription desc{Kind::Intrinsic, Intrinsic::opcode};       \
+    return desc;                                                               \
+  }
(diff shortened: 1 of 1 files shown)


lib/Dialect/OpDescription.cpp
```diff
@@ -116,215 +116,103 @@ template <> const OpDescription &OpDescription::get<BinaryOperator>() {
...
-  static unsigned opcodes[] = {
-      Intrinsic::dbg_declare, Intrinsic::dbg_value, Intrinsic::dbg_label,
-      Intrinsic::dbg_assign,
-      // Intrinsic::dbg_addr, <-- add this back for sufficiently recent LLVM
-  };
-  static const OpDescription desc{Kind::Intrinsic, opcodes};
-  return desc;
-}
+HANDLE_INTRINSIC_DESC_ARR(LifetimeIntrinsic, Intrinsic::lifetime_start,
+                          Intrinsic::lifetime_end)
```
nhaehnle, line 141: The HANDLE_xxx_INST that were already there have a special meaning in connection to including Instruction.def. Please do not change them.

Keep the definition of the intrinsic-related #defines separate and below the comment.

nhaehnle (commented):
Makes sense, but please see inline

tsymalla:
Added dedicated macros.

nhaehnle (approved):
Thanks!

PR #46, opened 2023-05-19
Add support for variadic arguments.

tsymalla:
This patch introduces a new binding varargs.
With varargs, an input argument can be bound to a list of arguments appearing at the tail of an operation.


tsymalla:
This is in a WIP state, so open for discussion and suggestions

lib/TableGen/Operations.cpp
```diff
@@ -570,24 +626,48 @@ void BuilderMethod::emitDefinition(raw_ostream &out, FmtContext &fmt,
     }
 
     fmt.addSubst("args", symbols.chooseName({"args", "arguments"}));
-    out << tgfmt("::llvm::Value* const $args[] = {\n", &fmt);
+    out << tgfmt("std::vector<::llvm::Value*> $args = {\n", &fmt);
```
Flakebi, line 629: Maybe use a SmallVector and use the number of arguments (1 for varargs) as the stack size?

lib/TableGen/Operations.cpp
```diff
@@ -570,24 +626,48 @@ void BuilderMethod::emitDefinition(raw_ostream &out, FmtContext &fmt,
...
+
+      ++ArgIdx;
     }
 
+    std::string varArgInitializer = "";
+    if (varArg) {
+      varArgInitializer = tgfmt(R"(
+        ::llvm::for_each($0, [&](::llvm::Value *varArg) {
+          $args.push_back(varArg);
+        });
```
Flakebi, line 659: For a SmallVector, this can be `$args.append($0.begin(), $0.end())`, which reserves the needed size upfront.

Flakebi (commented):
Nice!

tsymalla:
Added assertions, used SmallVector with fixed number of preserved elements, fixed wrong iterator range.

tsymalla:
Made iterator range for varargs explicit.

lib/TableGen/Operations.cpp
```diff
@@ -105,19 +108,32 @@ SmallVector<NamedValue> OperationBase::getFullArguments() const {
 }
 
 unsigned OperationBase::getNumFullArguments() const {
+  size_t argSize = m_arguments.size();
+
+  // Variadic argument lists can be of size zero,
+  // so do not add them to the total number of arguments.
+  if (m_hasVariadicArgument && !m_arguments.empty())
+    --argSize;
```
jasilvanus, line 116: Can the condition `m_hasVariadicArgument && m_arguments.empty()` (not the second condition is inverted) be true, and if so, in which cases?

lib/TableGen/Operations.cpp
```diff
@@ -127,41 +143,61 @@ void OperationBase::emitArgumentAccessorDefinitions(llvm::raw_ostream &out,
...
+
+    out << tgfmt(R"(
+      $cppType $_op::get$Name() {
+        return $fromLlvm;
+      })",
+                 &fmt);
+
+    // We do not generate a setter for variadic arguments for now.
+    if (arg.type->isVarArgList())
+      continue;
```
jasilvanus, line 184: Maybe split this into `generateGetter()` and `generateSetter()`? Then, this loop reduces to

```
generateGetter();
if (!arg.type->isVarArgList())
  generateSetter();
```

which may be a bit clearer.

lib/TableGen/Operations.cpp
tsymalla: Theoretically, it could, if you pass zero arguments and use a varargs binding without any arguments (e. g. `create<Op>({})`). In that case, the expression should evaluate to zero. Good catch. Thanks.

lib/TableGen/Operations.cpp
tsymalla: Will clean up, thanks.

lib/TableGen/Operations.cpp
tsymalla: From a runtime perspective (e. g. checking against concrete operations), yes. So, that would be a good catch.
From a dialects-compilation perspective, no, as using a variadic argument binding implies having an entry in `m_arguments`.
So, maybe the check to `!m_arguments.empty()` is superfluous, so I'll probably remove it. Thanks.

jasilvanus: Maybe put it into an assert.

tsymalla: I think we can leave it for now.

include/llvm-dialects/TableGen/Operations.h
```diff
@@ -141,6 +165,13 @@ class BuilderMethod {
   struct Arg {
     std::string cppType;
     std::string name;
+    std::optional<std::string_view> defaultValue;
+
+    void emitArg(llvm::raw_ostream &os) const {
+      os << cppType << " " << name;
+      if (defaultValue.has_value())
+        os << " = " << defaultValue.value();
+    }
```
nhaehnle, line 174: I don't think we should be adding these default values by default just because it's varargs and there's a simple default to give.

I'm not opposed to the idea of defaults for builder arguments, but it should be something that is chosen explicitly in the .td file. (In particular, there could also be a default for Attr* arguments.)

nhaehnle (commented):
This is already looking pretty good. I have a bunch of inline comments, and please add a check in the "parser" that if the superclass has varargs, we don't add any additional arguments on top.

include/llvm-dialects/TableGen/Operations.h
tsymalla: The idea is not to choose an explicit default in general, but to avoid having the user to specify the variadic argument in create calls in case he doesn't need them. So, for varargs, we always want to default to `{}`.
Separating the `Arg` and `BuilderArg` however makes it more complex to emit a default value based on the kind of the argument.
When going with the approach of defining the argument defaults in the dialect, we could create a custom DAG node for varargs where {} always is the default, but is there really any use case for defining other default values at this stage?

tsymalla:
Implemented most of the review comments. For the `superclass` approach, there's currently no link to the superclass of a record except the one in the `parseArguments` - so, for now, I'm just passing the information whether the superclass has a variadic argument to the parser. It might be cleaner to actually query the type hierarchy in the parser itself, but this requires additional changes. For now, the parser bails out if a situation occurs to notify the user something's wrong with the definitions.

Added a new test to show the behavior.

include/llvm-dialects/TableGen/Operations.h
tsymalla: If we can live with having the caller pass an empty array, that'd be fine from my side, and I'll remove the defaults again.

nhaehnle:
> For the superclass approach, there's currently no link to the superclass of a record except the one in the parseArguments

You could add the check in `OperationBase::init`.

tsymalla:
> > For the superclass approach, there's currently no link to the superclass of a record except the one in the parseArguments
> 
> You could add the check in `OperationBase::init`.
> 
> (Edit: The reason for doing that is that this is the one place that applies to both ops and opclasses but **not** other users of the underlying parsing method. Fewer boolean flags are always a good thing :))

But that would imply that we'd always (in case) successfully parse all arguments and not error out as early as we can.
But I'm fine with that idea as well, will update the PR later.

tsymalla:
Removed default values for builder args, so we now rely on the user. Added another example class WriteVarArgOp to simplify the tests a bit.

tsymalla:
Moved superclass varargs check to `OperationBase::init`, bails out asap.

lib/TableGen/Operations.cpp
```diff
@@ -64,11 +84,24 @@ bool OperationBase::init(raw_ostream &errs, GenDialectsContext &context,
...
+        errs << "Superclass " << m_superclass->name
+             << " already has variadic arguments, cannot set "
+                "again on "
+             << arg.name << ".\n";
+        return false;
+      }
+
+      m_hasVariadicArgument = true;
+    }
+
```
nhaehnle, line 104: If the superclass ends with varargs, then having *any* argument is not okay. The check should mostly be moved outside the loop; if the superclass has varargs, check that there aren't *any* arguments here, and also set this->m_hasVariadicArgument.

nhaehnle (commented):
Thanks! One more thing.

nhaehnle (approved):
Thanks!

lib/TableGen/Operations.cpp
tsymalla: Ah, I see, this is already handled by `SymbolTable::chooseName` - it seems to iteratively lookup the symbol table for either "args" or "arguments" and either chooses the other name or uses an incremental suffix. Cool! :)

PR #50, opened 2023-05-31
Remove unused capture in ConstraintSystem

tsymalla:
Fixes a warning during compilation.


nhaehnle (commented):
I don't understand why the clang-format check doesn't trigger, but this looks like bad formatting to me.

nhaehnle (approved):

PR #54, opened 2023-06-16
Fix vararg getter

nhaehnle:
arg_begin() and arg_end() return iterators over Use objects that are stored contiguously in memory, but the Value objects that the uses point to are very much not contiguous.

We solve this by introducing an iterator adaptor. The example code is also adjusted to test this in a better way.

### DIFF
--- a/include/llvm-dialects/Dialect/Utils.h
+++ b/include/llvm-dialects/Dialect/Utils.h
@@ -18,7 +18,9 @@
 
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/STLFunctionalExtras.h"
+#include "llvm/ADT/iterator.h"
 #include "llvm/IR/Type.h"
+#include "llvm/IR/User.h"
 #include "llvm/Support/raw_ostream.h"
 
 namespace llvm {
--- a/lib/TableGen/Constraints.cpp
+++ b/lib/TableGen/Constraints.cpp
@@ -312,7 +312,7 @@ StringRef MetaType::getGetterCppType() const {
     return "::llvm::Value *";
 
   if (isVarArgList())
-    return "::llvm::iterator_range<::llvm::Value *>";
+    return "::llvm::iterator_range<::llvm::User::value_op_iterator>";
 
   return getCppType();
 }
--- a/test/example/generated/ExampleDialect.cpp.inc
+++ b/test/example/generated/ExampleDialect.cpp.inc
@@ -1671,8 +1671,10 @@ data
       void WriteVarArgOp::setData(::llvm::Value * data) {
         setArgOperand(0, data);
       }
-      ::llvm::iterator_range<::llvm::Value *> WriteVarArgOp::getArgs() {
-        return ::llvm::make_range((arg_begin() + 1)->get(), (arg_begin() + arg_size() - 1)->get());
+      ::llvm::iterator_range<::llvm::User::value_op_iterator> WriteVarArgOp::getArgs() {
+        return ::llvm::make_range(
+            value_op_iterator(arg_begin() + 1),
+            value_op_iterator(arg_end()));
       }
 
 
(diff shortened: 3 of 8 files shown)


test/example/visitor-basic.ll
```diff
@@ -20,7 +22,7 @@ entry:
   %v2 = call i32 @llvm.umax.i32(i32 %v1, i32 %q)
   %t = call i8 (...) @xd.itrunc.i8(i32 %v2)
   call void (...) @xd.write(i8 %t)
-  call void (...) @xd.write.vararg(i8 %t, i32 %v2)
+  call void (...) @xd.write.vararg(i8 %t, i32 %v2, i32 %q)
```
jasilvanus, line 25: This example would have crashed before this fix?

test/example/generated/ExampleDialect.cpp.inc
```diff
@@ -1671,8 +1670,10 @@ data
       void WriteVarArgOp::setData(::llvm::Value * data) {
         setArgOperand(0, data);
       }
-      ::llvm::iterator_range<::llvm::Value *> WriteVarArgOp::getArgs() {
-        return ::llvm::make_range((arg_begin() + 1)->get(), (arg_begin() + arg_size() - 1)->get());
```
jasilvanus, line 1675: BTW, apart from the memory layout issue: Wouldn't the old version have missed the last element, passing it as `end`?

include/llvm-dialects/Dialect/Utils.h
```diff
@@ -42,6 +44,30 @@ bool runInstructionVerifier(
...
+          typename std::iterator_traits<BaseT>::iterator_category,
+          llvm::Value *> {
+private:
+  using SelfT = op_to_value_iterator_adaptor_impl<BaseT>;
+
+public:
+  op_to_value_iterator_adaptor_impl() = default;
+  template <typename U>
+  op_to_value_iterator_adaptor_impl(U &&u)
+      : SelfT::iterator_adaptor_base(std::forward<U &&>(u)) {}
```
jasilvanus, line 61: Do default copy and move constructor work together with this forwarding one being present, without having explicitly been declared?

tsymalla:
LGTM

lib/TableGen/Constraints.cpp
```diff
@@ -312,7 +312,8 @@ StringRef MetaType::getGetterCppType() const {
     return "::llvm::Value *";
 
   if (isVarArgList())
-    return "::llvm::iterator_range<::llvm::Value *>";
+    return "::llvm::iterator_range<::llvm_dialects::const_op_to_value_iterator_"
```
ruiling, line 315: Can we just return llvm::iterator_range<llvm::Use *> here? LLVM already provide an adaptor iterator to make iterating over the Values easy. See here: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/User.h#L250.

test/example/generated/ExampleDialect.cpp.inc
tsymalla: No, otherwise the tests would have failed.

nhaehnle: Yes, that was also wrong. The tests just didn't capture that, which is why I changed the visitor test.

include/llvm-dialects/Dialect/Utils.h
nhaehnle: I think so. At least, LLVM's pointee_iterator is defined using the same pattern.

test/example/visitor-basic.ll
nhaehnle: Yes, assuming the additional change in the visitor code.

lib/TableGen/Constraints.cpp
nhaehnle: Thanks -- I thought there must be something like this somewhere but couldn't find it :)

I do think this should return an iteration over Values and not Uses. It's technically less powerful, but it's less surprising given the better symmetry between what you put into the builder vs. what you get from the getter. I'll look into simplifying the change using the `User::value_op_iterator` next week.

test/example/generated/ExampleDialect.cpp.inc
tsymalla: What is the reason the test in test-builder.test did not fail then? Is this now working because of your change?

nhaehnle: Well, the code in ExampleMain.cpp was also wrong. It directly used the `end()` of the range, which should have failed because the end is supposed to be exclusive / one-past the range. But the bug in how `end` was set and the bug in how `end` was used canceled each other out.

PR #55, opened 2023-06-22
Include <variant>

jayfoad:
After LLVM patch https://reviews.llvm.org/D150997, `<variant>` is no longer included transitively here.


jasilvanus (approved):

jayfoad:
The shadercache builds failed with "SPIRV-Headers was not found - please checkout a copy under external/" so I'm going to ignore them.

PR #56, opened 2023-07-07
Add const to match{Declaration,Instruction} function arguments

jasilvanus:
Logically, this operation should not need to change the passed object (Instruction or Function), and indeed the implementation needs only `const` access.

### DIFF
--- a/include/llvm-dialects/Dialect/OpDescription.h
+++ b/include/llvm-dialects/Dialect/OpDescription.h
@@ -52,8 +52,8 @@ public:
   Kind getKind() const { return m_kind; }
   llvm::ArrayRef<unsigned> getOpcodes() const;
 
-  bool matchInstruction(llvm::Instruction &inst) const;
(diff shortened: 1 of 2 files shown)


jasilvanus:
Apparently something is wrong with the ubsan jobs:

```
error: Unable to fully sync the tree
error: Checking out local projects failed.
Failing repos:
drivers/llpc/imported/llvm-dialects
Try re-running with "-j1 --fail-fast" to exit at the first error.
The command '/bin/bash -c /vulkandriver/update-dialects.sh' returned a non-zero code: 1
```

Flakebi (approved):

Flakebi:
LGTM as most CI passed. The sanitizer jobs are a problem in the llpc repo… (it provides the base-builds for the CI here)

nhaehnle:
Agreed. I'm overriding the CI results for this one.

PR #57, opened 2023-07-13
Include StringExtras.h

dstutt:
After LLVM patch https://reviews.llvm.org/D155018, StringExtras.h is no longer included transitively.


jayfoad (approved):

dstutt:
I just wanted to highlight that the change here isn't adding an include into the cpp file - but rather is altering the headers that are generated (not obvious I think).

PR #59, opened 2023-07-21
Remove typed pointers from example

Flakebi:
Upstream llvm now asserts when trying to enable typed pointers.

llvm-dialects needs to keep support for typed pointers for some time, but it also needs to compile with upstream llvm.

We could add a CMake flag or make the test code dependent on the llvm version, but in practice, llvm-dialects examples are only tested here.


nhaehnle (approved):
Ugh, that's inconvenient. The better we can remove the dependency on typed pointers in all internal users the better. But yeah, this is probably the way to go here unless we wanted to add separate internal testing and jump through REQUIRES hoops or something.

PR #60, opened 2023-07-26
Remove use of deprecated Opaque Pointer API

mariusz-sikora-at-amd:
'isOpaque' and 'getNonOpaquePointerElementType' are deprecated on new LLVM. Additionally this code is not used since Opaque Pointers are enabled by default.

### DIFF
--- a/lib/Dialect/Utils.cpp
+++ b/lib/Dialect/Utils.cpp
@@ -50,11 +50,14 @@ static std::string getMangledTypeStr(Type *Ty, bool &HasUnnamedType) {
   std::string Result;
   if (PointerType *PTyp = dyn_cast<PointerType>(Ty)) {
     Result += "p" + utostr(PTyp->getAddressSpace());
+#if HAVE_LLVM_VERSION_MAJOR <= 17
     // Opaque pointer doesn't have pointee type information, so we just mangle
     // address space for opaque pointer.
+    // After llvm-18+ opaque pointer API will be removed from LLVM.
     if (!PTyp->isOpaque())
       Result += getMangledTypeStr(PTyp->getNonOpaquePointerElementType(),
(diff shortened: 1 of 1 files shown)


Flakebi (changes requested):
We cannot remove typed pointer support from llvm-dialects for now, this would break internal projects.
If we do need to remove this for new llvm, I think our only choice is to add `#if`s on the LLVM revision.

mariusz-sikora-at-amd:
> We cannot remove typed pointer support from llvm-dialects for now, this would break internal projects. If we do need to remove this for new llvm, I think our only choice is to add `#if`s on the LLVM revision.

Thanks, I was not aware about it. I only noticed this change: https://reviews.llvm.org/D155585
So, maybe we should add #if LLVM_VERSION_MAJOR <= 17 ?

Flakebi:
That check sounds good to me.
Thanks for doing all these changes!

mariusz-sikora-at-amd:
> That check sounds good to me. Thanks for doing all these changes!

Sure, no problem. I also noticed that we have a dedicated flag for opaque pointers in llvm-dialects
```
cl::opt<bool> g_typedPointers("typed-pointers", cl::init(false),
                               cl::desc("Disable opaque pointers"));
```
this will be also "useless" after llvm-17 ?

Flakebi:
Do we still have that flag somewhere? I thought I removed it in https://github.com/GPUOpen-Drivers/llvm-dialects/pull/59

mariusz-sikora-at-amd:
> Do we still have that flag somewhere? I thought I removed it in #59

I was referring exactly to this flag, but apparently I'm not sync with top.

Flakebi (approved):
Thanks!

PR #62, opened 2023-08-05
tablegen: relax a restriction of logic or

nhaehnle:
The constraint system evaluator refuses to evaluate a logic or unless
all variables that appear in it have been resolved. This is a good
thing, because we can't reliably capture variables from inside a logic
or.

However, the check was too restrictive since it also refused to evaluate
a logic or that contain a local/internal variable that is introduced
purely as a connective between "apply" constraints that appear as nested
tablegen DAGs.

We relax this restriction by making the distinction between "global" and
"non-global" variables explicit.

### DIFF
--- a/lib/TableGen/Constraints.cpp
+++ b/lib/TableGen/Constraints.cpp
@@ -64,6 +64,16 @@ Variable *Scope::createVariable(StringRef name, Predicate *predicate) {
 void ConstraintSystem::print(raw_ostream &out, StringRef prefix) const {
   for (const auto &constraintPtr : m_constraints)
     constraintPtr->print(out, prefix);
+
+  out << prefix << "free variables:";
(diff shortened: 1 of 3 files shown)


lib/TableGen/Constraints.cpp
```diff
@@ -118,8 +128,11 @@ bool ConstraintSystem::addConstraintImpl(raw_ostream &errs, Init *init,
       result->m_init = init;
       result->m_self = self;
 
-      if (self)
+      if (self && dag->getNumArgs()) {
```
tsymalla, line 131: `!dag->arg_empty()` ?

tsymalla (changes requested):

tsymalla:
LGTM

tsymalla-AMD (approved):

PR #63, opened 2023-08-09
Simplify operating on sets and maps of dialect ops.

tsymalla:
This commit introduces a new type, `OpSet`, that simplifies membership tests of an instruction on a set of dialect operations. With this PR, applying a callback on a given set of operations also becomes simpler since a visitor function `addSet` is introduced.

It also introduces a map-like data structure, `OpMap`, which can be used to simplify handling of association of dialect operations to certain values.

With this PR, a new check-llvm-dialects-units target is added which tests the new GTest ADT tests as well.


example/ExampleMain.cpp
```diff
@@ -4,28 +4,31 @@
...
  *
  *  Permission is hereby granted, free of charge, to any person obtaining a copy
- *  of this software and associated documentation files (the "Software"), to deal
- *  in the Software without restriction, including without limitation the rights
- *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
- *  copies of the Software, and to permit persons to whom the Software is
+ *  of this software and associated documentation files (the "Software"), to
+ *deal in the Software without restriction, including without limitation the
+ *rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ *sell copies of the Software, and to permit persons to whom the Software is
```
nhaehnle, line 10: Unrelated changes.

example/ExampleMain.cpp
```diff
@@ -166,12 +172,33 @@ LLVM_DIALECTS_VISITOR_PAYLOAD_PROJECT_FIELD(VisitorNest, inner)
...
+                    *self.out << "visiting SetWriteOp (set): " << op << '\n';
+                  }
+                });
+            b.addSet(complexSet, [](VisitorNest &self, llvm::Instruction &op) {
+              if (op.getOpcode() == Instruction::Ret) {
+                *self.out << "visiting Ret (set): " << op << '\n';
+              } else if (auto *II = dyn_cast<IntrinsicInst>(&op)) {
+                if (II->getIntrinsicID() == Intrinsic::umin) {
+                  *self.out << "visiting umin (set): " << op << '\n';
+                }
```
nhaehnle, line 199: Rephrase the checks here as assertions, since the use of `complexSet` should make these a guarantee. We don't want to accidentally miss a bug here.

include/llvm-dialects/Dialect/Dialect.h
```diff
@@ -249,7 +249,9 @@ bool isSimpleOperationDecl(const llvm::Function *fn, llvm::StringRef name);
 bool isOverloadedOperationDecl(const llvm::Function *fn, llvm::StringRef name);
 
 bool isSimpleOperation(const llvm::CallInst *i, llvm::StringRef name);
+bool isSimpleOperation(llvm::StringRef fn, llvm::StringRef name);
 bool isOverloadedOperation(const llvm::CallInst *i, llvm::StringRef name);
+bool isOverloadedOperation(llvm::StringRef fn, llvm::StringRef name);
```
nhaehnle, line 254: I don't think you need these overloads. Instead, add an `isOperationDecl` (see above) that takes an `isOverloaded` argument in addition to the name.

include/llvm-dialects/Dialect/Utils.h
```diff
@@ -78,4 +78,7 @@ template <typename T> Printable<T> printable(const T &x) {
   return Printable<T>(x);
 }
 
+bool isMatchingFunc(llvm::StringRef refFn, bool isOverload,
+                    llvm::StringRef mnemonic);
```
nhaehnle, line 82: See the other comment -- I don't think this is needed, just an `isOperationDecl` is fine and fits better to what we already have.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,356 @@
...
+
+  // -------------------------------------------------------------
+  // Comparison operator overloads.
+  // -------------------------------------------------------------
+  bool operator==(const OpMap &rhs) const {
+    return m_coreOpcodes == rhs.m_coreOpcodes &&
+           m_intrinsics == rhs.m_intrinsics && m_dialectOps == rhs.m_dialectOps;
+  }
+
+  bool operator!=(const OpMap &rhs) const { return !(*this == rhs); }
```
nhaehnle, line 75: I don't think these `operator==/!=` implement the correct semantics, since they depend on the ordering of the dialect ops vector. Do we need them in the first place?

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,356 @@
...
+
+  // -------------------------------------------------------------
+  // contains checks if a given op is contained in any of the
+  // internal map containers.
+  // -------------------------------------------------------------
+
+  // Check if the map contains either a core operation or an intrinsic.
+  bool contains(unsigned op) const {
```
nhaehnle, line 83: This needs to be split into two methods (or removed), because a core op code can alias an intrinsic ID.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,356 @@
...
+
+  // Return a reference to the operation described by a given OpDescription.
+  // This includes any of the possible operation kinds.
+  // Constructs a new entry if the operation was found in neither map.
+  ValueT &get(const OpDescription &desc) {
+    if (desc.isCoreOp()) {
+      for (const unsigned op : desc.getOpcodes()) {
+        if (m_coreOpcodes.contains(op))
+          return m_coreOpcodes[op];
+      }
```
nhaehnle, line 127: I think this loop, which has a dependency on the hash function used in `DenseMap`, proves that the `OpMap` class should refuse `OpDescription`s of multiple core opcodes and intrinsics in most or even all places.

This would also allow you to directly forward this to the correct `DenseMap` instead of forwarding to `insert_or_get` where a redundant kind check happens (which the compiler can probably optimize away if it inlines everything, but still).

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,356 @@
...
+    // Nothing found. Insert a new entry in the correct map.
+    return insert_or_get(desc, {});
+  }
+
+  // -------------------------------------------------------------
+  // Convenience operator[] definitions.
+  // -------------------------------------------------------------
+  ValueT &operator[](const IntrinsicInst &inst) {
+    return m_intrinsics[inst.getIntrinsicID()];
+  }
```
nhaehnle, line 153: It's weirdly non-orthogonal to have this overload but not one that works for general instructions. I know that we just can't have the latter, but I think this means we shouldn't have this overload either.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,356 @@
...
+  // containers.
+  // It returns either a default-constructed object if the key
+  // was not found or a copy of the contained value.
+  // -------------------------------------------------------------
+
+  template <typename OpT> ValueT lookup(const Function &func) const {
+    return lookup(OpDescription::get<OpT>(), func);
+  }
+
+  ValueT lookup(const OpDescription &desc, const Function &func) const {
```
nhaehnle, line 169: What is this and similar overloads good for? If we already had an `OpDescription`, then we could just lookup based on that, couldn't we? I suspect they should be removed.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    return {};
+  }
+
+  // Try to lookup an instruction which is either an intrinsic instruction,
+  // a dialect operation or a core instruction.
+  ValueT lookup(const Instruction &inst) const {
+    if (isa<IntrinsicInst>(inst) || isa<CallInst>(inst))
+      return lookup(*cast<CallInst>(inst).getCalledFunction());
```
nhaehnle, line 223: This can just be `isa<CallInst>`, since `IntrinsicInst` is derived from `CallInst`.

I wonder whether this case should fallback to looking up `OpDescription::get<CallInst>()`. I tend to think that it should, though in any case it feels likely that users of this functionality will always need some amount of special-casing around `CallInst`.

Note that this also needs to handle the case that `getCalledFunction` returns nullptr.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,356 @@
...
+
+    if (desc.isCoreOp()) {
+      for (const unsigned op : desc.getOpcodes())
+        erased |= m_coreOpcodes.erase(op);
+    } else if (desc.isIntrinsic()) {
+      for (const unsigned op : desc.getOpcodes())
+        erased |= m_intrinsics.erase(op);
+    }
+
+    // Try to erase the dialect op at last.
```
nhaehnle, line 320: This should be under an `else` (or just always directly `return` from the cases above)

test/unit/dialect/TestDialect.td
```diff
@@ -0,0 +1,73 @@
...
+
+class TestOp<string mnemonic_, list<Trait> traits_>
+    : Op<TestDialect, mnemonic_, traits_>;
+
+def DialectOp1 : TestOp<"dialect.op.1",
+                       []> {
+  let results = (outs);
+  let arguments = (ins);
+
+  let defaultBuilderHasExplicitResultType = true;
```
nhaehnle, line 41: That doesn't make a lot of sense when there's no result value in the first place :)

I think this test dialect needs at least two ops, one with and one without overloads.

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,267 @@
...
+    static OpDescription desc = OpDescription::get<OpT>();
+    return matches(desc, inst);
+  }
+
+  // Checks if `inst` is described by `desc`, if it is stored in the
+  // set.
+  bool matches(const OpDescription &desc, const Instruction &inst) const {
+    if (isa<IntrinsicInst>(inst) || isa<CallInst>(inst)) {
+      return matches(desc, *cast<CallInst>(inst).getCalledFunction());
+    }
```
nhaehnle, line 145: I notice there aren't any unit tests for `matches`. And similar to those analogous `lookup` overloads in `OpMap`, I don't understand what it's good for.

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,267 @@
...
+
+    if (isa<IntrinsicInst>(&Inst) || isa<CallInst>(&Inst))
+      return isMember(*cast<CallInst>(Inst).getCalledFunction());
+
+    return false;
+  }
+
+  // Checks if `inst` does not belong to the OpSet.
+  bool isNoMember(const Instruction &inst) const { return !isMember(inst); }
```
nhaehnle, line 195: Why?

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,267 @@
...
+    return contains(desc) && desc.matchDeclaration(func);
+  }
+
+  // -------------------------------------------------------------
+  // Membership tests check whether a given instruction or function is
+  // stored in the OpSet.
+  // -------------------------------------------------------------
+
+  // Checks if `inst` belongs to the OpSet.
+  bool isMember(const Instruction &Inst) const {
```
nhaehnle, line 182: Why not make this an overload of `contains`. That feels a bit more intuitive to me.

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,267 @@
...
+        if (func.getIntrinsicID() == Intrinsic)
+          return true;
+      }
+    }
+
+    return isMatchingDialectOp(func.getName());
+  }
+
+  // Checks if `func` does not belong to the OpSet.
+  bool isNoMember(const Function &func) const { return !isMember(func); }
```
nhaehnle, line 210: Why?

nhaehnle (commented):
Thank you, I like that you were quite thorough with the interface of OpSet and OpMap. I think you went a bit overboard in some places, but comments are inline.

I still need to take a closer look at the visitor parts.

include/llvm-dialects/Dialect/OpMap.h
tsymalla: Right now, unfortunately yes, otherwise the comparison tests don't work (see the `OpMapIRTests`).
You are right on the dependency of the ordering. Maybe it makes sense to always sort the vector once it was mutated.

include/llvm-dialects/Dialect/OpMap.h
tsymalla: Makes sense, thanks.

test/unit/dialect/TestDialect.td
tsymalla: Hm, the `DialectOp4` is used to show the intended behavior when a function is overloaded. Here, the same function is used with an `i32` and a `float` argument. This can be seen in the `DialectOpOverloadTests` - but I can extend the testing, if you want.

tsymalla:
Regenerated copyright comments.
Interestingly, local clang-format (16.0.6) breaks the comments for me.

tsymalla:
Side note - I did not work on all comments yet, just the obvious ones.

nhaehnle (commented):
The intention for the visitor implementation was to merge the set into the existing data structures, or perhaps just replace the existing data structures by an `OpMap<SmallVector<unsigned>>`. This is both for cleanliness and for performance reasons.

tsymalla:
Makes sense, will take care of it.

include/llvm-dialects/Dialect/OpMap.h
tsymalla: I just went with the approach to do a quick pairwise lookup, which should work fine.

tsymalla:
Update to address the remaining change requests.

I thought about the Visitor changes a bit more. Right now, in my opinion it only makes sense to either use the `OpMap` approach or to handle the set cases separately.
The issue with the current Visitor implementation is that for `OpDescription` objects, it relies on having a single instance obtained by OpDescription::get<OpT>(), which is static in the generated dialect. This works since for each `OpT`, there is a separate instance in the according `::get` method.
However, if we were to construct an `OpDescription` from the information stored in the `OpSet::m_dialectOps` object, we'd construct a `VisitorKey`-local object in some helper method for each dialect op in set. The Visitor needs to reference that since it doesn't own the `OpDescription` objects. This will lead to dangling pointers. However, making it static inside this `VisitorKey` method would also not work since all keys would use the same memory location.

Making the VisitorTemplate own the objects is also no appropriate solution in terms of perf / storage, so we'd rather get rid of the OpDescription approach in the `VisitorTemplate`. Happy for opinions.

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,227 @@
...
+
+    return isMatchingDialectOp(desc.getMnemonic());
+  }
+
+  // Checks if `inst` belongs to the OpSet.
+  bool contains(const Instruction &inst) const {
+    for (const unsigned CoreOp : m_coreOpcodes) {
+      if (inst.getOpcode() == CoreOp)
+        return true;
+    }
```
nhaehnle, line 145: For performance, this should make use of the fact that `m_coreOpcodes` is a set ;)

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,227 @@
...
+          return true;
+      }
+    }
+
+    if (desc.isIntrinsic()) {
+      for (const unsigned op : desc.getOpcodes()) {
+        if (containsIntrinsicID(op))
+          return true;
+      }
+    }
```
nhaehnle, line 135: I still think this should assert that there is only a single opcode / intrinsic ID in the OpDescription.

nhaehnle:
This sounds like using an `OpMap` is the way to go.

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  // Checks if `func` belongs to the OpSet.
+  bool contains(const Function &func) const {
+    if (func.isIntrinsic()) {
+      for (const unsigned Intrinsic : m_intrinsicIDs) {
+        if (func.getIntrinsicID() == Intrinsic)
+          return true;
+      }
+    }
```
nhaehnle, line 166: m_intrinsicIDs is a set. And I think this can early-out by returning `false` here.

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,227 @@
...
+
+    if (desc.isIntrinsic()) {
+      for (const unsigned op : desc.getOpcodes()) {
+        if (containsIntrinsicID(op))
+          return true;
+      }
+    }
+
+    return isMatchingDialectOp(desc.getMnemonic());
```
nhaehnle, line 137: Can't this compare the OpDescription directly?

include/llvm-dialects/Dialect/OpSet.h
```diff
@@ -0,0 +1,227 @@
...
+
+    return false;
+  }
+
+  // Tries to insert a given description in the internal data structures.
+  void tryInsertOp(const OpDescription &desc) {
+    if (desc.isCoreOp()) {
+      for (const unsigned op : desc.getOpcodes())
+        m_coreOpcodes.insert(op);
+    }
```
nhaehnle, line 207: Early-out here and below.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,357 @@
...
+    static OpDescription desc = OpDescription::get<OpT>();
+    return contains(desc);
+  }
+
+  // Check if the map contains an op described by an OpDescription.
+  bool contains(const OpDescription &desc) const {
+    if (desc.isCoreOp() || desc.isIntrinsic()) {
+      for (const unsigned op : desc.getOpcodes()) {
+        if (containsCoreOp(op) || containsIntrinsic(op))
+          return true;
```
nhaehnle, line 125: This needs to only check either `containsCoreOp` or `containsIntrinsic`, depending on the incoming `desc`.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,357 @@
...
+    return {};
+  }
+
+  // Try to lookup an instruction which is either an intrinsic instruction,
+  // a dialect operation or a core instruction.
+  ValueT lookup(const Instruction &inst) const {
+    if (auto *CI = dyn_cast<CallInst>(&inst)) {
+      const Function *Callee = CI->getCalledFunction();
+      if (!Callee)
+        return {};
```
nhaehnle, line 210: This case could still fall through to the core opcode case, since `Instruction::Call` could be in `m_coreOpcodes`.

include/llvm-dialects/Dialect/Dialect.h
```diff
@@ -245,11 +245,14 @@ class DialectExtensionRegistration {
...
-bool isSimpleOperation(const llvm::CallInst *i, llvm::StringRef name);
-bool isOverloadedOperation(const llvm::CallInst *i, llvm::StringRef name);
+bool isSimpleOperationDecl(const llvm::Function *fn, llvm::StringRef mnemonic);
+bool isOverloadedOperationDecl(const llvm::Function *fn,
+                               llvm::StringRef mnemonic);
+
+bool isSimpleOperation(const llvm::CallInst *i, llvm::StringRef mnemonic);
+bool isOverloadedOperation(const llvm::CallInst *i, llvm::StringRef mnemonic);
+bool isOperationDecl(llvm::StringRef fn, bool isOverloaded,
+                     llvm::StringRef mnemonic);
```
nhaehnle, line 255: I still think this would be better served with a Function argument for consistency. I think the one place it's used without a function in OpSet could be solved like in OpMap instead.

nhaehnle:
Interesting, yes. I think the "key" part of the iterator should just be an OpDescription? That would be consist with the "base" interface for looking things up in the map.

tsymalla:
Right, this is the general idea I have (plus some additional state to keep track of what we are actually looking at).
However, this requires to add a default constructor for the `OpDescription`, which we don't want to expose as public API, otherwise we cannot initialize storage in the constructor. We could befriend the `OpMap` to the `OpDescription`, which is ugly as well.
I am not certain it is really that bad we add an default `OpDescription` constructor, since we never do any checks in the remaining `OpDescription` constructors, so we could initialize that with arbitrary (invalid) state as well.

tsymalla:
> Right, this is the general idea I have (plus some additional state to keep track of what we are actually looking at). However, this requires to add a default constructor for the `OpDescription`, which we don't want to expose as public API, otherwise we cannot initialize storage in the constructor. We could befriend the `OpMap` to the `OpDescription`, which is ugly as well. I am not certain it is really that bad we add an default `OpDescription` constructor, since we never do any checks in the remaining `OpDescription` constructors, so we could initialize that with arbitrary (invalid) state as well.

(we could derive the iterator from `std::pair<OpDescription, ValueT *>` as well, but that's also ugly)

tsymalla:
Added initial refactoring stub using an `OpMapIterator`. This works, but I am quite unhappy with the current solution, so open for suggestions.

First, it is not easy to write an iterator around iterators of two different types. I observed this issue since the beginning. Second, storing an offset into the internal containers seems to work, but it is quite ugly. Storing pointers to the `OpMap` data structures internally seems not to be simply doable.
Third, the current approach uses an O(n) loop for every `operator++`  invocation in the core operations and intrinsic operations state, since it uses `std::advance` on the `DenseMapIterator`. I'd rather pass in an pointer to the previously locally declared iterator and increment that, if it's not `nullptr`, but that requires some ugly workarounds to make it work as well. So, I am leaving that for review at the moment and am very happy for suggestions. Rebase will follow in a sec.

nhaehnle:
> Storing iterators to the OpMap data structures internally seems not to be simply doable (since we need to express the constness as defined in the DenseMapIterator in the OpMapIterator as well).

Yeah, being generic over `const`-ness is basically not possible in C++.

The canonical trick here is to define a base CRTP class along the lines of:
```
  template <typename IterT, typename ValueT, typename MapIterator, typename DialectOpIterator>
  struct base_iterator {
  private:
    struct TaggedCoreIterator : MapIterator { /* probably need constructors */ };
    struct TaggedIntrinsicIterator : MapIterator { /* probably need constructors */ };

    const OpMapT *m_map;
    std::variant<DialectOpIterator, CoreIterator, IntrinsicIterator> m_it;

  public:
    ...
  };
```
and then instantiate as
```
  struct const_iterator : base_iterator<iterator, const ValueT, DenseMap<unsigned, ValueT>::const_iterator, SmallVectorImpl<...>::const_iterator> {
    ... will need at least some constructors here ...
  };

  // and then analogously for iterator
```

`operator*` can perhaps just return a `std::pair<OpDescription, ValueT &>` by value. In that case `operator->` has to return a wrapper type.

nhaehnle (commented):
> Storing iterators to the OpMap data structures internally seems not to be simply doable (since we need to express the constness as defined in the DenseMapIterator in the OpMapIterator as well).

Yeah, being generic over `const`-ness is basically not possible in C++.

The canonical trick here is to define a base CRTP class along the lines of:
```
  template <typename IterT, typename ValueT, typename MapIterator, typename DialectOpIterator>
  struct base_iterator {
  private:
    struct TaggedCoreIterator : MapIterator { /* probably need constructors */ };
    struct TaggedIntrinsicIterator : MapIterator { /* probably need constructors */ };

    const OpMapT *m_map;
    std::variant<DialectOpIterator, CoreIterator, IntrinsicIterator> m_it;

  public:
    ...
  };
```
and then instantiate as
```
  struct const_iterator : base_iterator<iterator, const ValueT, DenseMap<unsigned, ValueT>::const_iterator, SmallVectorImpl<...>::const_iterator> {
    ... will need at least some constructors here ...
  };

  // and then analogously for iterator
```

`operator*` can perhaps just return a `std::pair<OpDescription, ValueT &>` by value. In that case `operator->` has to return a wrapper type.

tsymalla:
I am sorry the next update took a little bit longer, this was because of known reasons.

I implemented the const- and non-const version via `std::conditional_t`, which helps integrating everything in a single class, without inheritance.
This required a little bit of fiddling around with the function members from `OpMap`. I ultimatively decided to make the `OpMapIterator` smart, which means to move all relevant logic to generate the right iterator to the `OpMap` internals to the `OpMapIterator`.
Plus, the `Visitor` now should fully support the `OpMap`, and for me, offline, it passes. I see some failures on CI which seem to have to do with repo? Maybe something is wrong, which needs to be fixed in `main` as well - since there has been no commit for quite a while now, we'd test that out.

Performance-wise, this indeed is a bit of overhead, namingly the state-machine approach when iterating through the `OpMap`. I also added a bit of preprocessor "magic" to not have all the cumbersome `iterator` / `const_iterator` overloads inside the `OpMap`...

Happy for comments.

tsymalla:
The ubsan+asan Jobs complain:
_30.19 info: A new version of repo is available
30.19 warning: repo is not tracking a remote branch, so it will not receive updates
30.19 repo reset: error: Entry '.flake8' not uptodate. Cannot merge.
30.19 fatal: Could not reset index file to revision 'v2.36.1^0'.
30.19 
------
dialects.Dockerfile:34
--------------------
  32 |     
  33 |     # Sync the repos. Replace the base repo with a freshly checked-out one.
  34 | >>> RUN /vulkandriver/update-dialects.sh
  35 |     
  36 |     # Build llvm-dialects._

Flakebi:
The CI needs to be updated to not use the shadercache anymore. I’ll take a look at that.

Flakebi:
Should be fixed in #65

tsymalla-AMD:
Thanks, @Flakebi !
Rebased this PR.

tsymalla-AMD:
Fix issue caught by UBSan

tsymalla-AMD:
Remove unused member

tsymalla-AMD:
Change: Hide `OpMapIterator` constructors from other clients, since its functionality should only be exposed via the `OpMap`.

tsymalla:
Fixed iteration bug when invoking `begin()` with empty core op container.
Added iterator test.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,814 @@
...
+
+  // -------------------------------------------------------------
+  // Convenience operator[] definitions.
+  // -------------------------------------------------------------
+
+  ValueT &operator[](const OpDescription &desc) {
+    return findOrConstruct(desc);
+  }
+  ValueT *operator[](const Function &func) { return findPointer(func); }
+  ValueT *operator[](const Instruction &inst) { return findPointer(inst); }
```
nhaehnle, line 268: Remove the `operator[]` overloads that return a pointer. That's too surprising.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -0,0 +1,814 @@
...
+
+    return {};
+  }
+
+  // -------------------------------------------------------------
+  // Inserts a given Key/Value pair and return a reference or
+  // directly return a reference if the key already exists.
+  // -------------------------------------------------------------
+
+  ValueT &findOrConstruct(const OpDescription &desc, const ValueT &val) {
```
nhaehnle, line 303: This API is too surprising. When you say "find or construct" I hear "either find (and leave the map unmodified) or construct a new element". What this method does instead is to unconditionally assign `val` to `desc`.

It seems this is primarily required for the copy constructor. Could you instead just implement `try_emplace`? That is:
```c++
  template <typename... Ts>
  std::pair<iterator, bool> try_emplace(const OpDescription &desc, Ts &&... args) {
    ...
    auto [it, inserted] = m_coreOpcodes.try_emplace(op, std::forward<Ts>(args)...);
    return {iterator_t(this, it, false), inserted};
    ...
  }
```

include/llvm-dialects/Dialect/Visitor.h
```diff
@@ -287,9 +300,8 @@ class VisitorBase {
   VisitorStrategy m_strategy;
   std::vector<PayloadProjection> m_projections;
   std::vector<VisitorHandler> m_handlers;
-  llvm::DenseMap<unsigned, HandlerRange> m_coreOpcodeMap;
-  llvm::DenseMap<unsigned, HandlerRange> m_intrinsicIdMap;
-  std::vector<std::pair<const OpDescription *, HandlerRange>> m_dialectCases;
+  DenseMap<unsigned, HandlerRange> m_map2;
```
nhaehnle, line 303: This looks like an leftover from an earlier version?

lib/Dialect/Visitor.cpp
```diff
@@ -51,37 +51,58 @@ void VisitorTemplate::add(VisitorKey key, VisitorCallback *fn,
...
+    m_opMap[OpDescription::fromIntrinsic(key.m_intrinsicId)].push_back(
+        handlerIdx);
+  } else if (key.m_kind == VisitorKey::Kind::OpDescription) {
     const OpDescription *description = key.m_description;
     switch (description->getKind()) {
     case OpDescription::Kind::Core:
-      for (unsigned opcode : description->getOpcodes())
-        m_coreOpcodeMap[opcode].push_back(handlerIdx);
+      for (unsigned op : description->getOpcodes())
+        m_opMap[OpDescription::fromCoreOp(op)].push_back(handlerIdx);
```
nhaehnle, line 67: It should be possible to simplify this entire block of code down to `m_opMap[*description].push_back(handleIdx);`

lib/Dialect/Visitor.cpp
```diff
@@ -51,37 +51,58 @@ void VisitorTemplate::add(VisitorKey key, VisitorCallback *fn,
...
+      const OpDescription desc{dialectOpPair.isOverload,
+                               dialectOpPair.mnemonic};
+
+      if (m_opMap.contains(desc)) {
+        m_opMap[desc].push_back(handlerIdx);
+      } else {
+        SmallVector<unsigned> handlers;
+        handlers.push_back(handlerIdx);
+        m_opMap[desc] = std::move(handlers);
+      }
```
nhaehnle, line 104: This can just be `m_opMap[desc].push_back(handlerIdx)`

nhaehnle (commented):
Sorry it took me so long to get around to it. The iterator solution looks pretty good now and I think it's really quite close. Though there are some interface issues that I think should really be addressed before we commit it. As a general rule, the interfaces of OpMap and OpSet should be as unsurprising and as close to the corresponding STL classes where possible. There are some gaps there still, see inline.

include/llvm-dialects/Dialect/OpMap.h
tsymalla: Yeah, makes sense. This requires some adjustments. Generally, an issue for the Visitor refactoring is that we constrained the `OpDescriptions` to be inserted to be single-opcodes. This means, something like that:

```
      m_opMap.insert(try_emplace, {handlerIdx});
```

does not neccessarily work.

I'd rather remove that constraint for insertion and return an iterator range instead. Same goes for the `OpMapIterator::erase()` method, since it is rather inefficient to only erase a single operation at a time.

nhaehnle: Hmm, but if the OpDescription has multiple opcodes, then inserted locations don't necessarily fall into a contiguous range of iterators.

For the longer term, I'd rather remove the support for multi-opcodes/multi-intrinsics in OpDescription. It's always been more of a hack.

tsymalla:
Addressed review comments.
Renamed tests.
Implemented try_emplace iterator-based approach.
Fixed issue with new try_emplace-based approach, where the iterator would compare DenseMap iterators built in different epochs.
Added some more tests to demonstrate functionality when trying to mix different kinds of operations in the OpMap.

nhaehnle (approved):
Thanks! While I haven't gone over it again with a comb, I think the overall design is solid and we can continue to iterate if anything comes up.

tsymalla:
Two minor changes: Removed a superfluous code line from `OpSet` and replaced a call to `OpMap::insert` in the Visitor with a call to `OpMap::operator[]` for consistency reasons. Should not make any difference and I'll merge after the checks have passed.

PR #66, opened 2023-10-12
OpMap: Fix StateMachine issue.

tsymalla-AMD:
When we are in the CoreOp or Intrinsic state and are about to transition to the DialectOp state, then we cannot just access the DialectOpIteratorT in the internal variant storage, since the new iterator (m_dialectOps.begin()) is not stored in the variant yet.


nhaehnle (approved):
LGTM

PR #73, opened 2023-11-29
Add support for typed pointers with the builtin PointerType.

tsymalla:
In some scenarios, we want to use typed pointers even if LLVM is moving to opaque pointers. Dialects currently only supports opaque pointers in its builtin. With this change, we simply return an i8* when evaluating a builtin pointer type return value or operand. This translates to an opaque pointer as well.
The client that creates the operations is responsible for handling the pointers correctly (e. g. not using opaque pointers in a
typed pointer scenario).

### DIFF
--- a/include/llvm-dialects/Dialect/Dialect.td
+++ b/include/llvm-dialects/Dialect/Dialect.td
@@ -169,7 +169,8 @@ def I64 : TgConstant<(IntegerType 64)>, Type;
 
 def PointerType : BuiltinType {
   let arguments = (args type:$self, AttrI32:$address_space);
-  let evaluate = "::llvm::PointerType::get($_context, $address_space)";
+  // Supports both typed and opaque pointers.
+  let evaluate = "::llvm::PointerType::get(::llvm::Type::getInt8Ty($_context), $address_space)";
   let check = "$self->isPointerTy()";
   let capture = ["$self->getPointerAddressSpace()"];
 }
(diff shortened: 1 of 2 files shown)


include/llvm-dialects/Dialect/Dialect.td
```diff
@@ -169,7 +169,7 @@ def I64 : TgConstant<(IntegerType 64)>, Type;
 
 def PointerType : BuiltinType {
   let arguments = (args type:$self, AttrI32:$address_space);
-  let evaluate = "::llvm::PointerType::get($_context, $address_space)";
+  let evaluate = "($_context.supportsTypedPointers() ? ::llvm::PointerType::get(::llvm::Type::getInt8Ty($_context), $address_space) : ::llvm::PointerType::get($_context, $address_space))";
```
tsymalla-AMD, line 172: Probably can just return an Int8PtrTy, since this will convert to opaque pointers in an opaque pointer scenario anyway?

Flakebi: Yes, that should make all of this shorter :)

Flakebi (commented):
LGTM apart from a nit inline

example/ExampleMain.cpp
```diff
@@ -253,6 +253,7 @@ int main(int argc, char **argv) {
   llvm::cl::ParseCommandLineOptions(argc, argv);
 
   LLVMContext context;
+  context.setOpaquePointers(true);
```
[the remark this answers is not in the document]
tsymalla: Yeah, but it will fail tests without if we run check-llvm-dialects on internal repos

PR #74, opened 2023-12-12
Remove using namespace

Flakebi:
Using namespace in headers is prone to generate conflicts, so specify the namespace explicitly.


jasilvanus (approved):
LGTM if tests are passing.

nhaehnle (approved):

PR #78, opened 2024-01-09
Fix llvm-dialects-tblgen Android corss-compiling issue

mdinkov:
Compilation of intermediate llvm-dialects-tblgen currently requires the Android linker to be located /system/bin/linker*. Editing the root directory can be disallowed on some build systems. This change allows the Android linker to be fetched directly from the AOSP build (i.e. without having to copy it to /system/bin/)

Change authored by Jinchao Xu.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,6 @@ endif()
 target_link_libraries(llvm-dialects-tblgen PRIVATE llvm_dialects_tablegen)
 set_compiler_options(llvm-dialects-tblgen)
 
-#if __ANDROID__
 if(ANDROID)
     if (ANDROID_ABI MATCHES "x86_64")
         target_link_options(llvm-dialects-tblgen PRIVATE
@@ -97,7 +96,6 @@ if(ANDROID)
         )
     endif()
 endif()
-#endif
 
 # The llvm_dialects library build depends on llvm/IR/Attributes.inc
 add_dependencies(llvm_dialects intrinsics_gen)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,23 @@ endif()
 target_link_libraries(llvm-dialects-tblgen PRIVATE llvm_dialects_tablegen)
 set_compiler_options(llvm-dialects-tblgen)
 
+#if __ANDROID__
+if(ANDROID)
+    if (ANDROID_ABI MATCHES "x86_64")
+        target_link_options(llvm-dialects-tblgen PRIVATE
+            LINKER:--rpath,$ENV{ANDROID_PRODUCT_OUT}/system/lib64/bootstrap/:$ENV{ANDROID_PRODUCT_OUT}/system/lib64/
+            LINKER:-dynamic-linker,$ENV{ANDROID_PRODUCT_OUT}/apex/com.android.runtime/bin/linker64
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -82,6 +82,23 @@ endif()
 target_link_libraries(llvm-dialects-tblgen PRIVATE llvm_dialects_tablegen)
 set_compiler_options(llvm-dialects-tblgen)
 
+#if __ANDROID__
```
jasilvanus, line 85: Looks like a code sanitization comment not needed on GPU Open?

CMakeLists.txt
```diff
@@ -82,6 +82,23 @@ endif()
...
+            LINKER:--rpath,$ENV{ANDROID_PRODUCT_OUT}/system/lib64/bootstrap/:$ENV{ANDROID_PRODUCT_OUT}/system/lib64/
+            LINKER:-dynamic-linker,$ENV{ANDROID_PRODUCT_OUT}/apex/com.android.runtime/bin/linker64
+        )
+    elseif(ANDROID_ABI MATCHES "x86")
+        # There is a limitation here. The PID should less than 65535 for 32-bit executable.
+        target_link_options(llvm-dialects-tblgen PRIVATE
+            LINKER:--rpath,$ENV{ANDROID_PRODUCT_OUT}/system/lib/bootstrap/:$ENV{ANDROID_PRODUCT_OUT}/system/lib/
+            LINKER:-dynamic-linker,$ENV{ANDROID_PRODUCT_OUT}/apex/com.android.runtime/bin/linker
+        )
+    endif()
```
jasilvanus, line 98: I know very little about cross compilation, but it seems a bit weird to teach a generic tool like LLVM dialects about specifics of Android builds. How does that work for other tools, e.g. as part of LLVM?

Also, depending on environment variables feels a bit weird -- aren't these available as CMake variables?

mdinkov: To my knowledge, this is the only place where we've had to use the Android linker (which typically gets installed on the Android system itself). The initial solution was to copy the linker to /system/bin/linker(64), as that's where the linker is expected on the Android system. However, copying Android binaries (and libraries) into the root system of the Linux build machine is hacky.

I don't believe we have a CMake variable necessarily pointing to the out directory, as this is the only item that has to be retrieved from the AOSP build output. The remaining AOSP dependencies can be fetched from the prebuilt VNDK.

trenouf:
Shouldn't this tie into, or at least be analogous to, the way the NATIVE build works on llvm? That's where tablegen gets built in llvm. I guess mlir probably does the same.

nhaehnle:
> Shouldn't this tie into, or at least be analogous to, the way the NATIVE build works on llvm? That's where tablegen gets built in llvm. I guess mlir probably does the same.

Agreed. I'm not familiar with how this works, but a quick skim suggests re-using / adapting the pattern in `llvm/cmake/modules/TableGen.cmake` where `build_native_tool` is invoked.

For some added context, I initially didn't just use the `add_tablegen` macro because it only worked from inside an LLVM build, and the intention was to make it possible to use llvm-dialects with a pre-built version of LLVM. I believe there is some interest with Julia folks to potentially do that. But perhaps for a cross-compilation, making some assumptions doesn't hurt.

PR #79, opened 2024-01-12
Make Visitor more robust on instruction iteration

xuechen417:
Use a fixed end guard for the instruction iteration is not robust for callbacks to add a basic block. Use `Instruction::getNextNode()` for the iteration to allow callbacks to directly edit the code.


xuechen417:
Move `getNextNode` before calling `visit`

xuechen417:
ping for review

nhaehnle (approved):
Thanks!

PR #82, opened 2024-01-31
Merge the lit and unit test suite.

tsymalla:
Rename `check-llvm-dialects` => `check-llvm-dialects-lit` Add a new custom target `check-llvm-dialects` that runs both `check-llvm-dialects-lit` and `check-llvm-dialects-units`.

Run the combined test suite on CI.

While we only use the units test suite for `OpMap` and `OpSet` right now, it makes sense to have them executed both under a common name.

### DIFF
--- a/docker/dialects.Dockerfile
+++ b/docker/dialects.Dockerfile
@@ -38,10 +38,7 @@ WORKDIR /vulkandriver/builds/ci-build
 RUN source /vulkandriver/env.sh \
     && cmake --build . --target llvm-dialects-tblgen
 
-# Run the lit test suite.
+# Run all test suites.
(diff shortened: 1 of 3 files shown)


docker/dialects.Dockerfile
```diff
@@ -38,10 +38,7 @@ WORKDIR /vulkandriver/builds/ci-build
 RUN source /vulkandriver/env.sh \
     && cmake --build . --target llvm-dialects-tblgen
 
-# Run the lit test suite.
+# Run all test suite.
```
JanRehders-AMD, line 41: suite -> suites

nhaehnle (approved):
Thanks

PR #83, opened 2024-01-31
Change name mangling for intrinsic functions fixing #2

JanRehders-AMD:
We want to avoid confusion when one op's intrinsic is the prefix of another one. Using a string `__` which we don't use for intrinsic function names ensures this. Using "$" was considered but `__` was chosen so the LLVM IR printer will not add quotes to those names which would make updating the tests much harder.


JanRehders-AMD:
@nhaehnle @tsymalla @Flakebi

lib/Dialect/Dialect.cpp
```diff
@@ -281,7 +281,8 @@ bool llvm_dialects::detail::isOperationDecl(llvm::StringRef fn,
     if (!fn.starts_with(mnemonic))
       return false;
 
-    return fn[mnemonic.size()] == '.';
+    return fn.size() + 2 >= mnemonic.size() && fn[mnemonic.size()] == '_' &&
```
nhaehnle, line 284: Shouldn't this be `fn.size() > mnemonic.size() + 2`? (Strict > makes sense because something has to be after the underscores, though it doesn't *really* matter)

nhaehnle (commented):
One issue, rest LGTM

lib/Dialect/Dialect.cpp
JanRehders-AMD: Agree that it doesn't matter much. But before something like lgc.foobar. was resulted in returning true, so I'd err on the side of caution here and also let this match for lgc.foobar__ now

tsymalla, line 284: Can't you just use `.substr()`?

JanRehders-AMD: The "__" will not be at the end of the intrinsic name. And substr would make a copy and allocate

tsymalla: Hm, yeah, I missed that this will not be at the end of the intrinsic name.

nhaehnle (approved):

PR #85, opened 2024-02-01
Remove trailing CR in test-generated.ll.

tsymalla:
Add the option `strip-trailing-cr` to avoid CR/LF-related test failures. The dialect files are created with LF, thus we don't want to see diff mismatches in git environments which check out with CRLF.


nhaehnle (approved):
Seems reasonable to me.

PR #88, opened 2024-02-05
Error if duplicated mnemonics are found.

tsymalla:
We currently don't emit an error when a dialect defines two dialect operations with the same mnemonic. This change emits basic information about duplicate mnemonics.

Since this requires a `O(n)` loop worst-case, store the mnemonic counts per dialect in a `StringMap` and iterate over it during finalization of the dialect, which then also bails out.

Can possibly be extended to all kinds of duplicated symbols.

### DIFF
--- a/lib/TableGen/Operations.cpp
+++ b/lib/TableGen/Operations.cpp
@@ -447,6 +447,7 @@ bool Operation::parse(raw_ostream &errs, GenDialectsContext *context,
 
   op->m_builders.push_back(std::move(builder));
 
+  ++dialect->operationCounts[op->mnemonic];
   dialect->operations.push_back(std::move(op));
 
   return true;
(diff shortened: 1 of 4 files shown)


lib/TableGen/Operations.cpp
```diff
@@ -447,6 +447,7 @@ bool Operation::parse(raw_ostream &errs, GenDialectsContext *context,
 
   op->m_builders.push_back(std::move(builder));
 
+  ++dialect->operationCounts[op->mnemonic];
```
Flakebi, line 450: How about using a `StringSet`?
You can try inserting `op->mnemonic` into the set here. If it was already in the set (returned bool is `false`), throw an error about a duplicate definition.

tsymalla: My initial idea was to just do a std::adjacent_find on the final sorted list of operations (in ::finalize), but I felt it might be more helpful if all duplicate ops are listed at once.
Since `Operation::parse` is only for a single op,I wanted to leave the actual validation work to `GenDialect`.

But honestly, I am unemotional about the final approach.

Flakebi: Yeah, I although thought about that after I left the comment. Fine with me to leave it as is.

nhaehnle (approved):

PR #91, opened 2024-05-21
Add support for immutable attributes.

tsymalla:
Sometimes we don't want or can't update operands with a auto-generated setter method, so this change adds support to make attributes immutable.

### DIFF
--- a/include/llvm-dialects/Dialect/Dialect.td
+++ b/include/llvm-dialects/Dialect/Dialect.td
@@ -69,6 +69,9 @@ class Attr<string cppType_> : MetaType {
   // A check statement that is issued before using the C++ value in builders.
   // $0 is the C++ value.
   string check = "";
+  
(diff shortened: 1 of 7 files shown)


include/llvm-dialects/Dialect/Dialect.td
```diff
@@ -69,6 +69,9 @@ class Attr<string cppType_> : MetaType {
   // A check statement that is issued before using the C++ value in builders.
   // $0 is the C++ value.
   string check = "";
+  
+  // Overriding prevents generating a setter method. Attributes are mutable by default.
+  string isImmutable = "false";
```
nhaehnle, line 74: TableGen has booleans...
```suggestion
  bit isImmutable = false;
```
... and knock-on changes that are required for that.

nhaehnle (commented):
One kind of obvious remark :)

nhaehnle (approved):
Thanks!

PR #92, opened 2024-05-21
Add support for immutable strings.

tsymalla:
We want a way to store strings in a global variable by passing in a
StringRef to an op, but we don't want to generate a setter for it, since
we currently don't inject the builder into the setter. So, add an
immutable string type based on the existing isImmutable option for
attributes.

Based on #91


include/llvm-dialects/Dialect/Dialect.td
```diff
@@ -279,6 +279,21 @@ def : AttrLlvmType<AttrI16, I16>;
...
+}
+ 
+def ImmutableStringType : BuiltinType {
+  let arguments = (args type:$self);
+  let evaluate = "::llvm::PointerType::get($_context, 0)";
+  let check = "$self->isPointerTy()";
+  let capture = [];
+}
+ 
+def : AttrLlvmType<ImmutableStringAttr, ImmutableStringType>;
```
nhaehnle, line 295: I would have thought this could just use `Ptr` instead of a new type. Does that not work?

example/ExampleMain.cpp
```diff
@@ -141,6 +141,8 @@ void createFunctionExample(Module &module, const Twine &name) {
   moreVarArgs.push_back(b.getInt32(4));
   b.create<xd::InstNameConflictVarargsOp>(moreVarArgs, "four.varargs");
 
+  b.create<xd::StringAttrOp>("Hello world!");
```
nhaehnle, line 144: Please also add a test for the getter by visiting the StringAttrOp.

nhaehnle (commented):
Thanks! Just two comments on the relevant commit here.

nhaehnle (approved):
LGTM, thanks! (looks like the test failures are all covered by the discussionin #93)

PR #93, opened 2024-05-22
Fix argument count check for ops with only a variadic argument list.

tsymalla:
We don't count variadic arguments in the `getNumFullArguments()` method since they can be zero. Not properly checking the return value of this function can however lead to a `arg_size() < 0` check in the generated verifier method. This happens if we only have a single variadic argument list argument.

This could also be solved with a call to `std::max` but just let us not emit verifier code instead.


nhaehnle (approved):
Change looks good to me.

The sporadic test failure sounds like perhaps there's a missing dependency specification in the CMake.

PR #94, opened 2024-05-23
Improve lit testing scripts.

tsymalla:
- Let `check-llvm-dialects-lit` depend on the `llvm-dialects-test-depends` custom target
- Convert `add_dialects_unit_test` to a macro

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,7 +1,7 @@
 ##
  #######################################################################################################################
  #
- #  Copyright (c) 2020-2023 Advanced Micro Devices, Inc. All Rights Reserved.
+ #  Copyright (c) 2020-2024 Advanced Micro Devices, Inc. All Rights Reserved.
  #
  #  Permission is hereby granted, free of charge, to any person obtaining a copy
  #  of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -42,10 +42,9 @@ configure_lit_site_cfg(
 
 add_lit_testsuite(check-llvm-dialects-lit "Running the llvm-dialects regression tests"
   ${CMAKE_CURRENT_BINARY_DIR}
-  ${exclude_from_check_all}
-  DEPENDS ${LLVM_DIALECTS_TEST_DEPENDS}
+  ${EXCLUDE_FROM_CHECK_ALL}
+  DEPENDS llvm-dialects-test-depends
 )
-set_target_properties(check-llvm-dialects-lit PROPERTIES FOLDER "Tests")
```
nhaehnle, line 48: Why did you remove this? AFAIK it is relevant for Visual Studio.

nhaehnle (commented):
Not sure I understand what all the changes are supposed to do, but it seems reasonable enough overall.

test/CMakeLists.txt
tsymalla-AMD: It is already set on llvm-dialects-test-depends (not sure if it works though, going to double-check)

PR #95, opened 2024-06-06
Add `Dialect::isDialectOp` helper function.

tsymalla:
We sometimes want to check if a given call instruction is prefixed with the cpp namespace assigned to the dialect. Instead of doing that manually, add a static dialect function that does this check.

### DIFF
--- a/lib/TableGen/GenDialect.cpp
+++ b/lib/TableGen/GenDialect.cpp
@@ -107,6 +107,9 @@ class Builder;
 
     public:
       static Key& getKey();
+      static bool isDialectOp(::llvm::CallInst& op);
+      static bool isDialectOp(::llvm::Function& func);
+      static bool isDialectOp(::llvm::StringRef funcName);
 
     private:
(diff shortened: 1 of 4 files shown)


nhaehnle (approved):
LGTM

lib/TableGen/GenDialect.cpp
```diff
@@ -298,6 +301,18 @@ void llvm_dialects::genDialectDefs(raw_ostream& out, RecordKeeper& records) {
       return s_key;
     }
 
+    bool $Dialect::isDialectOp(::llvm::CallInst& op) {
+      return isDialectOp(op.getCalledFunction()->getName());
```
Flakebi, line 305: This should maybe check that `op.getCalledFunction() != nullptr` (which happens for indirect calls).

tsymalla: https://github.com/GPUOpen-Drivers/llvm-dialects/pull/96

PR #97, opened 2024-06-21
Add ImmutableString  default name of a global variable

jiaolu:
Add name to the global variable as the name is required for the crossliner to include as global variable


Flakebi (approved):
Sounds good to me

PR #101, opened 2024-07-17
Add functionality to re-create op with new variadic argument list.

tsymalla:
Since we cannot update varargs directly, and often need to re-create the op, add a small helper that re-creates a given op with a new variadic argument list and invalidates the current op.


tsymalla-AMD:
Not using the `create<>` method of the dialects builder here, since we cannot directly derive the correct types for the passed arguments (we not always pass in a `Value *`).

lib/TableGen/Operations.cpp
```diff
@@ -162,10 +163,18 @@ void OperationBase::emitArgumentAccessorDeclarations(llvm::raw_ostream &out,
...
-      )";
+    if (!arg.type->isImmutable()) {
+      if (!isVarArg) {
+        defaultDeclaration += R"(
+          void set$1($0 $3);
+        )";
+      } else {
+        defaultDeclaration += R"(
+          /// Returns a new op with the same arguments and a new tail argument list.
+          /// The object on which this is called will be invalidated.
```
Flakebi, line 174: nit: I’d say ‘replaced and erased’ instead of invalidated

lib/TableGen/Operations.cpp
```diff
@@ -162,10 +163,18 @@ void OperationBase::emitArgumentAccessorDeclarations(llvm::raw_ostream &out,
...
+          $_op *replace$1AndInvalidate(::llvm_dialects::Builder &, ::llvm::ArrayRef<Value *>);
```
Flakebi, line 175: I think this function should not take a builder and instead create the CallInst directly.
For a setter like this, it’s more convenient if one does not need to pass a builder.

lib/TableGen/Operations.cpp
```diff
@@ -247,6 +278,10 @@ void OperationBase::emitArgumentAccessorDefinitions(llvm::raw_ostream &out,
     fmt.addSubst("Name", convertToCamelFromSnakeCase(arg.name, true));
 
     builder.emitAccessorDefinitions();
+    if (!arg.type->isImmutable() && arg.type->isVarArgList())
+      builder.emitVarArgReplacementDefinition(numArgs);
```
Flakebi, line 282: I suppose `indexedArg.index()` is the same as `numArgs`

lib/TableGen/Operations.cpp
```diff
@@ -228,12 +237,34 @@ void AccessorBuilder::emitSetterDefinition() const {
                 &m_fmt);
 }
 
+void AccessorBuilder::emitVarArgReplacementDefinition(
+    const size_t numNonVarArgs) const {
+  std::string toLlvm = m_arg.name;
+
+  m_fmt.addSubst("numNonVarargs", std::to_string(numNonVarArgs));
```
Flakebi, line 244: It seems like `Twine(numNonVarArgs)` should work as well.

lib/TableGen/Operations.cpp
```diff
@@ -228,12 +237,34 @@ void AccessorBuilder::emitSetterDefinition() const {
...
+        ::llvm::SmallVector<Value *> newArgs;
+        if ($numNonVarargs > 0)
+          newArgs.append(arg_begin(), arg_begin() + $numNonVarargs);
+        newArgs.append($1.begin(), $1.end());
+        $_op *newOp = cast<$_op>(B.CreateCall(getCalledFunction(), newArgs, this->getName()));
+        this->replaceAllUsesWith(newOp);
+        this->eraseFromParent();
+        return newOp;
+      })",
+                &m_fmt, convertToCamelFromSnakeCase(toLlvm, true), toLlvm);
```
Flakebi, line 258: The toLlvm formatting arguments here are not needed, you can use `$name` instead of `$1`

lib/TableGen/Operations.cpp
tsymalla-AMD: I'm not certain on that, see the comment above. I am not sure we shouldn't pass a builder that guards and re-sets the insertion point.

Flakebi: I think using CallInst::Create is fine in this case

Flakebi (approved):
LGTM with the nits inline fixed

PR #103, opened 2024-07-25
Delete MMX name mangler case

Sisyph:
The type was removed in https://github.com/llvm/llvm-project/pull/98505


Sisyph:
@nhaehnle @dstutt @Flakebi

tsymalla-AMD (approved):

PR #105, opened 2024-10-02
Reflect upstream const changes

tsymalla:
`TableGenFnMain` was changed to pass a `const RecordKeeper &`, so
reflect that change in `llvm-dialects`. Also, `const Record *` pointers
are handed out now.
Adds a bunch of changes to make `Record *` const overall.
Guard both versions with LLVM_MAIN_REVISION to ensure upstream and
downstream CI continues to work.

### DIFF
--- a/include/llvm-dialects/TableGen/Common.h
+++ b/include/llvm-dialects/TableGen/Common.h
@@ -17,6 +17,15 @@
 #pragma once
 
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/TableGen/Record.h"
+
+#if !defined(LLVM_MAIN_REVISION) || LLVM_MAIN_REVISION >= 513628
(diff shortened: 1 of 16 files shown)


dstutt (approved):
As we discussed - this seems the most pragmatic approach.
We should probably remove this again at some point when the change has propagated everywhere it is required.
LGTM

jasilvanus:
Note we already have a dependency on `LLVM_MAIN_REVISION` elsewhere in `llvm-dialects` (although that one could likely be removed now).

tsymalla:
> > Note we already have a dependency on `LLVM_MAIN_REVISION` elsewhere in `llvm-dialects` (although that one could likely be removed now).
> 
> Yes, this single dependency can be removed now, I will take care of that later.

https://github.com/GPUOpen-Drivers/llvm-dialects/pull/107

PR #106, opened 2024-10-02
Fix namespace issue in `isDialectOp`

tsymalla:
This is using the `cppNamespace`. Use the IR-level namespace instead.


example/ExampleDialect.td
```diff
@@ -26,7 +26,7 @@
 include "llvm-dialects/Dialect/Dialect.td"
 
 def ExampleDialect : Dialect {
-  let name = "xd";
+  let name = "xd.ir";
   let cppNamespace = "xd";
```
jasilvanus, line 30: Let's also change the `cpp` namespace to clearly indicate this is the cpp namespace.

jasilvanus (approved):

PR #111, opened 2024-10-22
Comma expresion for MSVC fold expression bug

GinShio:
Boolean fold expression seems like some bugs in MSVC, it's incorrectly
removed. In other projects, we found similar problems. But gcc or clang
not.

On the other hand, fold expression with comma seems cleaner than lambda.

Referenced-by: zigrazor/cxxgraph#415


nhaehnle (approved):
Seems reasonable, thanks!

PR #112, opened 2024-10-23
Change all 'Init' pointers to const

mariusz-sikora-at-amd:
This change reflect upstream change:
https://github.com/llvm/llvm-project/pull/112705


tsymalla:
Looks good, thanks

Flakebi (approved):
Thanks!

PR #113, opened 2024-11-22
Add argument index enum per op

tsymalla:
Add an enum that helps identifying specific argument slots by providing access to the indices used in `getArgOperand`.

### DIFF
--- a/lib/TableGen/Operations.cpp
+++ b/lib/TableGen/Operations.cpp
@@ -21,6 +21,7 @@
 #include "llvm-dialects/TableGen/Format.h"
 
 #include "llvm/TableGen/Record.h"
+#include <map>
 
 using namespace llvm;
 using namespace llvm_dialects;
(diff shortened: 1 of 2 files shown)


lib/TableGen/Operations.cpp
```diff
@@ -159,8 +160,20 @@ unsigned OperationBase::getNumFullArguments() const {
 
 void OperationBase::emitArgumentAccessorDeclarations(llvm::raw_ostream &out,
                                                      FmtContext &fmt) const {
-  for (const auto &arg : m_arguments) {
+  std::map<std::string, uint32_t> argIndexMap;
```
Flakebi, line 163: Can this be a simple vector from 0 to `m_arguments.size()`?
Apart from being simpler, that should also fix the nondeterministic iteration order when printing the enum variants.

tsymalla-AMD: It cannot be `m_arguments.size()` since otherwise we reserve memory for the non-vararg arguments in the dialect op definition, leading to empty enum values, but otherwise it's a good idea, thanks.

jasilvanus (commented):
Thanks for working on this.

Maybe we should also add a `TailStart` or `VarArgsStart` value?

jasilvanus (commented):
Thanks, LGTM.

I'll leave approval to @Flakebi  or @nhaehnle.

lib/TableGen/Operations.cpp
```diff
@@ -159,8 +159,20 @@ unsigned OperationBase::getNumFullArguments() const {
...
 void OperationBase::emitArgumentAccessorDeclarations(llvm::raw_ostream &out,
                                                      FmtContext &fmt) const {
-  for (const auto &arg : m_arguments) {
+  llvm::SmallVector<std::string> argNames;
+
+  unsigned numSuperclassArgs = 0;
+  if (m_superclass)
+    numSuperclassArgs = m_superclass->getNumFullArguments();
+
+  for (const auto &[index, arg] : llvm::enumerate(m_arguments)) {
```
Flakebi, line 168: `index` is not used anymore

Flakebi (approved):
LGTM with a nit

PR #114, opened 2024-12-03
Use ArgumentIndex enum

tsymalla:
Instead of inserting the index value, use the enum value from ArgumentIndex.

### DIFF
--- a/test/example/generated/ExampleDialect.cpp.inc
+++ b/test/example/generated/ExampleDialect.cpp.inc
@@ -322,25 +322,25 @@ return true;
 
     
       ::llvm::Value * StreamReduceOp::getPtr() const {
-        return getArgOperand(0);
+        return getArgOperand(static_cast<uint32_t>(ArgumentIndex::Ptr));
       }
 
       void StreamReduceOp::setPtr(::llvm::Value * ptr) {
-        setArgOperand(0, ptr);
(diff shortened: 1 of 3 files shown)


test/example/generated/ExampleDialect.cpp.inc
```diff
@@ -322,25 +322,25 @@ return true;
 
     
       ::llvm::Value * StreamReduceOp::getPtr() const {
-        return getArgOperand(0);
+        return getArgOperand(static_cast<uint32_t>(ArgumentIndex::Ptr));
```
Flakebi, line 325: Now that I see all these static casts, maybe it would be better to define a non-class enum, which does not need explicit conversion to uint32_t:
```c++
// Define a non-class enum, so that it can be used as index without casts, but is still accessed like a class enum
struct ArgumentIndex {
enum : uint32_t {
  Ptr = 0,
  Count = 1,
};
};
```

jasilvanus: +1 on the idea to get rid of the casts, however, I think we should also make sure the enum type is accessible so it can be stored without auto, maybe like this?

```c++
struct ArgumentIndex {
enum Enum : uint32_t {
   Ptr = 0,
   Cont = 1
};
};
```

Flakebi (approved):
LGTM

PR #115, opened 2025-01-21
Allow Visitor to use CallInst

TylerNowicki:
* We have encountered IR where it is necessary to transform a CallInst.
* This requires a CallInst visitor. Although, a CallInst visitor can be
  added to the OpMap, it fails to find the CallInst visitor. Instead it
  gives up when it determines the Call is not an intrinsic or
  llvm-dialect op.
* This change ensures we can find the CallInst visitor.

### DIFF
--- a/include/llvm-dialects/Dialect/OpMap.h
+++ b/include/llvm-dialects/Dialect/OpMap.h
@@ -605,15 +605,12 @@ template <typename ValueT, bool isConst> class OpMapIteratorBase final {
       invalidate();
   }
 
-  // Do a lookup for a given instruction. Mark the iterator as invalid
-  // if the instruction is a call-like core instruction.
+  // Do a lookup for a given instruction.
   OpMapIteratorBase(OpMapT *map, const llvm::Instruction &inst) : m_map{map} {
     if (auto *CI = llvm::dyn_cast<llvm::CallInst>(&inst)) {
       const llvm::Function *callee = CI->getCalledFunction();
-      if (callee) {
-        if (createFromFunc(*callee))
-          return;
-      }
+      if (callee && createFromFunc(*callee))
+        return;
     }
 
     const unsigned op = inst.getOpcode();
--- a/include/llvm-dialects/Dialect/OpMap.h
+++ b/include/llvm-dialects/Dialect/OpMap.h
@@ -595,13 +595,14 @@ template <typename ValueT, bool isConst> class OpMapIteratorBase final {
         if (std::get<BaseIteratorT>(m_iterator) == map->m_intrinsics.end())
           invalidate();
       }
-    } else {
-      createFromDialectOp(desc.getMnemonic());
+    } else if (!createFromDialectOp(desc.getMnemonic())) {
+      invalidate();
     }
   }
 
   OpMapIteratorBase(OpMapT *map, const llvm::Function &func) : m_map{map} {
-    createFromFunc(func);
+    if (!createFromFunc(func))
+      invalidate();
   }
 
   // Do a lookup for a given instruction. Mark the iterator as invalid
@@ -610,18 +611,12 @@ template <typename ValueT, bool isConst> class OpMapIteratorBase final {
     if (auto *CI = llvm::dyn_cast<llvm::CallInst>(&inst)) {
       const llvm::Function *callee = CI->getCalledFunction();
       if (callee) {
-        createFromFunc(*callee);
-        return;
+        if (createFromFunc(*callee))
+          return;
       }
     }
 
     const unsigned op = inst.getOpcode();
-
-    // Construct an invalid iterator.
-    if (op == llvm::Instruction::Call || op == llvm::Instruction::CallBr) {
-      invalidate();
-      return;
-    }
 
     BaseIteratorT it = m_map->m_coreOpcodes.find(op);
     if (it != m_map->m_coreOpcodes.end()) {
@@ -699,20 +694,20 @@ protected:
 private:
   void invalidate() { m_isInvalid = true; }
 
-  void createFromFunc(const llvm::Function &func) {
+  bool createFromFunc(const llvm::Function &func) {
     if (func.isIntrinsic()) {
       m_iterator = m_map->m_intrinsics.find(func.getIntrinsicID());
 
       if (std::get<BaseIteratorT>(m_iterator) != m_map->m_intrinsics.end()) {
         m_desc = OpDescription::fromIntrinsic(func.getIntrinsicID());
-        return;
+        return true;
       }
     }
 
-    createFromDialectOp(func.getName());
+    return createFromDialectOp(func.getName());
   }
 
-  void createFromDialectOp(llvm::StringRef funcName) {
+  bool createFromDialectOp(llvm::StringRef funcName) {
     size_t idx = 0;
     bool found = false;
     for (auto &dialectOpKV : m_map->m_dialectOps) {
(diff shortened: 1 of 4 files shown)


include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -610,19 +611,13 @@ template <typename ValueT, bool isConst> class OpMapIteratorBase final {
     if (auto *CI = llvm::dyn_cast<llvm::CallInst>(&inst)) {
       const llvm::Function *callee = CI->getCalledFunction();
       if (callee) {
-        createFromFunc(*callee);
-        return;
+        if (createFromFunc(*callee))
```
tsymalla-AMD, line 614: Nit: Put this in the `if (callee)` branch above

test/unit/interface/OpMapIRTests.cpp
```diff
@@ -252,3 +253,43 @@ TEST_F(OpMapIRTestFixture, DialectOpOverloadTests) {
...
+  Builder.SetInsertPoint(getEntryBlock());
+
+  // Create a dummy global variable of type %OpaqueTy*
+  GlobalVariable *GV = new GlobalVariable(
+      *Mod, OpaqueTy, false, GlobalValue::PrivateLinkage, nullptr, "handle");
+  GV->setInitializer(ConstantAggregateZero::get(OpaqueTy));
+  Value *Op2 = GV;
+
+  // Create a constant value (e.g., 123)
+  Value *Op1 = ConstantInt::get(I32Ty, 123);
```
tsymalla-AMD, line 281: `Builder.getInt32()`?

tsymalla-AMD (changes requested):
Thanks!
Can you please extend the IR-level tests as well, so we can see it handles IR with a call and a callbr properly, when there is neither a dialect op nor an intrinsic callback registered?

test/unit/interface/OpMapIRTests.cpp
```diff
@@ -252,3 +253,43 @@ TEST_F(OpMapIRTestFixture, DialectOpOverloadTests) {
...
+TEST_F(OpMapIRTestFixture, CallInstCoreOpMatchesInstructionTest) {
+  // Declare %OpaqueTy = type opaque
+  StructType *OpaqueTy = StructType::create(Context, "OpaqueTy");
+
+  // Define types
+  PointerType *OpaquePtrTy = PointerType::get(OpaqueTy, 0);
+  IntegerType *I32Ty = Type::getInt32Ty(Context);
+
+  // Declare: %OpaqueTy* @ProcOpaqueHandle(i32, %OpaqueTy*)
+  FunctionType *ProcOpaqueHandleFuncTyTy =
```
Flakebi, line 266: TyTy? I think one Ty can be removed here (and below)

test/unit/interface/OpMapIRTests.cpp
```diff
@@ -252,3 +253,43 @@ TEST_F(OpMapIRTestFixture, DialectOpOverloadTests) {
   EXPECT_EQ(map.lookup(Op1), "DialectOp4");
   EXPECT_EQ(map.lookup(Op2), "DialectOp4");
 }
+
+TEST_F(OpMapIRTestFixture, CallInstCoreOpMatchesInstructionTest) {
+  // Declare %OpaqueTy = type opaque
+  StructType *OpaqueTy = StructType::create(Context, "OpaqueTy");
+
+  // Define types
+  PointerType *OpaquePtrTy = PointerType::get(OpaqueTy, 0);
```
Flakebi, line 262: Note that pointer types are opaque, so this is just a pointer. With a builder, it would be `Builder.getPtrTy()`.

include/llvm-dialects/Dialect/OpMap.h
```diff
@@ -610,19 +611,13 @@ template <typename ValueT, bool isConst> class OpMapIteratorBase final {
     if (auto *CI = llvm::dyn_cast<llvm::CallInst>(&inst)) {
```
Flakebi, line 611: Can you update the comment on this constructor?

Flakebi (commented):
Looks good, some nitpicks inline

test/unit/interface/OpMapIRTests.cpp
```diff
@@ -252,3 +253,43 @@ TEST_F(OpMapIRTestFixture, DialectOpOverloadTests) {
...
+  // Create a constant value (e.g., 123)
+  Value *Op1 = ConstantInt::get(I32Ty, 123);
+
+  // Build the call to ProcOpaqueHandle
+  Value *Args[] = {Op1, Op2};
+  const CallInst &CallInst = *Builder.CreateCall(ProcOpaqueHandleFuncTy, Args);
+
+  // Add Instruction::Call to OpMap
+  OpMap<StringRef> map;
+  const OpDescription CallDesc = OpDescription::fromCoreOp(Instruction::Call);
```
tsymalla-AMD, line 289: Can you please also test-case `::CallBr` here?

TylerNowicki:
> when there is neither a dialect op nor an intrinsic callback registered?

I don't understand this part of your request.

test/unit/interface/OpMapIRTests.cpp
```diff
@@ -252,3 +253,65 @@ TEST_F(OpMapIRTestFixture, DialectOpOverloadTests) {
...
+
+TEST_F(OpMapIRTestFixture, CallCoreOpMatchesInstructionTest) {
+  OpMap<StringRef> map;
+  llvm_dialects::Builder B{Context};
+
+  // Define types
+  PointerType *PtrTy = B.getPtrTy();
+  IntegerType *I32Ty = Type::getInt32Ty(Context);
+
+  // Declare: %ptr @ProcOpaqueHandle(i32, %ptr)
```
Flakebi, line 265: nitpick 😉
```suggestion
  // Declare: ptr @ProcOpaqueHandle(i32, ptr)
```

Flakebi (commented):
If I understand Thomas correctly, he means using a visitor with a `CallInst` in the example here: https://github.com/GPUOpen-Drivers/llvm-dialects/blob/dev/example/ExampleMain.cpp
I think that makes sense

TylerNowicki:
> If I understand Thomas correctly, he means using a visitor with a `CallInst` in the example here: https://github.com/GPUOpen-Drivers/llvm-dialects/blob/dev/example/ExampleMain.cpp I think that makes sense

At this point, from what I can see, we don't have any Visitor tests... it would help if there was an example I can work from.

TylerNowicki:
> > Oh, do you mean, add the CallInst to the example?
> 
> Yeah, I meant that. And then print something to `out` and check it in the test like for the intrinsics.

I guess its done... I don't really get it what this is doing.

Flakebi:
> I don't really get it what this is doing.

Ah, you need to add IR and checks here as well (and probably in the other test file): https://github.com/GPUOpen-Drivers/llvm-dialects/blob/dev/test/example/visitor-basic.ll

tsymalla-AMD:
Yeah. Sebastian understood correctly, please extend the `visitor-basic.ll` test + ExampleMain so that we can see the visitor visiting a generic CallInst.

tsymalla-AMD:
> I guess its done... I don't really get it what this is doing.

The Example app is invocated during testing. It runs the visitor declared there against a given input LL file and then filechecks against the output there - so it checks if the callbacks are properly invoked.

TylerNowicki:
Done

tsymalla:
Looks good to me

tsymalla-AMD (approved):

Flakebi (approved):
Nice!

PR #116, opened 2025-01-24
Add native llvm-dialects-tblgen for cross-compiling

robotchaoX:
[Description]
Add native llvm-dialects-tblgen for ARM64 Windows and Android cross-compiling. Use LLVM interface add_tablegen gengrate dialects TableGen executable. LLVM 17.0 rename "LLVM_DIALECTS-tablegen-host" to "llvm-dialects-tblgen-host" [LLVM][CMake] add_tablegen: Rename host tablegen from '${project}-tablegen-host' to '${target}-host' in LLVM 17.0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,21 @@ separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
 add_definitions(${LLVM_DEFINITIONS_LIST})
 add_definitions("-DHAVE_LLVM_VERSION_MAJOR=${LLVM_VERSION_MAJOR}")
 
+set(LLVM_DIALECTS_TABLEGEN_SOURCES
+    lib/TableGen/Common.cpp
+    lib/TableGen/Constraints.cpp
+    lib/TableGen/Dialects.cpp
+    lib/TableGen/DialectType.cpp
+    lib/TableGen/Evaluator.cpp
+    lib/TableGen/Format.cpp
+    lib/TableGen/GenDialect.cpp
+    lib/TableGen/NamedValue.cpp
+    lib/TableGen/Operations.cpp
+    lib/TableGen/Predicates.cpp
+    lib/TableGen/SymbolTable.cpp
+    lib/TableGen/Traits.cpp
+    )
+
 set(llvm_deps
     Support
     TableGen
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -63,8 +78,20 @@ if (llvm_dialects_is_in_llvm_build_tree)
         ${llvm_deps}
     )
 
-    set(LLVM_LINK_COMPONENTS ${llvm_deps})
-    add_llvm_tool(llvm-dialects-tblgen DISABLE_LLVM_LINK_LLVM_DYLIB)
+    set(PROJECT_NAME LLVM)
```
tsymalla-AMD, line 81: Why do you set this variable?

CMakeLists.txt
```diff
@@ -63,8 +78,20 @@ if (llvm_dialects_is_in_llvm_build_tree)
...
+    set(LLVM_LINK_COMPONENTS Support)
```
tsymalla-AMD, line 82: Support is already included in `llvm_deps`, why not change this variable?

robotchaoX: LLVM_LINK_COMPONENTS was used in add_tablegen explicitly. We should set `Support` explicitly, `TableGen` is there by default.
https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/TableGen.cmake#L179
```
macro(add_tablegen target project)
  cmake_parse_arguments(ADD_TABLEGEN "" "DESTINATION;EXPORT" "" ${ARGN})
  set(${target}_OLD_LLVM_LINK_COMPONENTS ${LLVM_LINK_COMPONENTS})
  set(LLVM_LINK_COMPONENTS ${LLVM_LINK_COMPONENTS} TableGen)
```

`llvm_deps` was used here too. So, I keep it.
https://github.com/GPUOpen-Drivers/llvm-dialects/blob/dev/CMakeLists.txt#L97

CMakeLists.txt
robotchaoX: `CONFIGURE_${PROJECT_NAME}_NATIVE` was defined in llvm-project `llvm_create_cross_target`
https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/CrossCompile.cmake#L109
```
  add_custom_target(CONFIGURE_${project_name}_${target_name}
    DEPENDS ${${project_name}_${target_name}_BUILD}/CMakeCache.txt)
```

Here `PROJECT_NAME` was used in llvm-project, it should be `LLVM`. 
If we not set it to `LLVM`, current project name llvm-dialects would cause issue.
https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/CrossCompile.cmake#L148
add_tablegen -> build_native_tool 
```
function(build_native_tool target output_path_var)
...
 add_custom_command(OUTPUT "${output_path}"
                     COMMAND ${build_cmd}
                     DEPENDS CONFIGURE_${PROJECT_NAME}_NATIVE ${ARG_DEPENDS}
                     WORKING_DIRECTORY "${${PROJECT_NAME}_NATIVE_BUILD}"
                     COMMENT "Building native ${target}..."
                     USES_TERMINAL)
```

tsymalla-AMD:
Looking okay to me, unfortunately we don't have CI testing for this specific feature.
But as long as it doesn't break existing builds, it should be okay.

nhaehnle (approved):

PR #118, opened 2025-02-17
Auto-generate comments for dialect ops.

tsymalla:
Currently, the summary and description fields for dialect ops remain unused. Add those as comment above the class declaration.


Flakebi (commented):
Nice, I have no comments on the code, just the CI seems to be failing on the generated example

nhaehnle:
Ugh... we shouldn't allow defining ops without summary or description :(

tsymalla-AMD:
> Ugh... we shouldn't allow defining ops without summary or description :(

I can validate that as follow-up.

PR #120, opened 2025-02-25
Argument attributes support

sstipano:
It would make sense to be able to mark some ops' arguments as NoCapture, (eg. `BufferLengthOp`).


sstipano:
I'm not able to add reviewers. @nhaehnle @tsymalla

sstipano:
Implemented in #121

PR #121, opened 2025-03-06
Value traits support

sstipano:


example/ExampleDialect.td
```diff
@@ -141,6 +141,13 @@ def Add32Op : ExampleOp<"add32", [Memory<[]>, NoUnwind, WillReturn]> {
   let results = (outs I32:$result);
   let arguments = (ins I32:$lhs, I32:$rhs, AttrI32:$extra);
 
+  let value_traits = [
+    (NoCapture $lhs),
+    (NoCapture $rhs),
```
nhaehnle, line 146: Having `nocapture` on an int argument is pretty strange. Best to add that to an op that has pointer types.

lib/TableGen/Operations.cpp
```diff
@@ -363,6 +363,47 @@ bool Operation::parse(raw_ostream &errs, GenDialectsContext *context,
   for (RecordTy *traitRec : record->getValueAsListOfDefs("traits"))
     op->traits.push_back(context->getTrait(traitRec));
 
+
+  const RecordVal *insVal = record->getValue("arguments");
+  std::unordered_map<std::string, unsigned> nameToIndexMap;
+  if (const DagInit *DI = dyn_cast<DagInit>(insVal->getValue())){
```
nhaehnle, line 369: It is an error for the `arguments` field to not be of dag type, and we should err on the side of failing loudly on unexpected data. Otherwise, users may end up specifying their operations incorrectly without noticing it.

Just use `getValueAsDag` here, as is done elsewhere. And the same applies below to `results` and analogously to `value_traits`.

lib/TableGen/Traits.cpp
```diff
@@ -52,12 +52,46 @@ class LlvmEnumAttributeTrait : public LlvmAttributeTrait {
...
 
+class LlvmEnumFnAttributeTrait : public LlvmEnumAttributeTrait {
+public:
+  LlvmEnumFnAttributeTrait()
+      : LlvmEnumAttributeTrait(Kind::LlvmEnumFnAttributeTrait) {}
+
+  void addAttribute(llvm::raw_ostream &out, FmtContext &fmt) const override;
+};
+class LlvmEnumRetAttributeTrait : public LlvmEnumAttributeTrait {
+public:
```
nhaehnle, line 63: Separate class declarations with a new line.

nhaehnle (commented):
Thank you for working on this!

One high-level observation is that this new feature should compose well with the `superclass` feature. In the example dialect, look at `StreamReduceOp`: it should be possible to add `value_traits` in there as well, which are then adopted by `StreamAddOp` etc. Please add that to the .td files as an example and test case.

On the C++ side, it's probably best to move the `traits` vector from `Operation` into `OperationBase`, from where it can be copied from the superclass in `OperationBase::init` if there is one.

Also... I may have talked about subclassing `LlvmEnumAttributeTrait` in C++, but after seeing that in code in practice I have to say that the redundancy between indices and classes is confusing. I think it would be best to stick with a single LlvmEnumAttributeTrait class with an `m_index` member, where the index uses magic numbers to indicate return value and function attributes, just like how it works in LLVM itself.

lib/TableGen/Operations.cpp
sstipano: Agree for the `arguments` field. Changed it to use `getValueAsDag`. However, `results` might be empty when working with `OpClass`es. For the `value_traits`, I added the `fatal_error_report` in case there is something other than dag inside of `value_traits` list.

nhaehnle: For the `results` case, you can still assume that `results` is a dag. That should even be ensured by the type system of the TableGen frontend. The only check you gneed is for `getNumArgs`.

nhaehnle (commented):
Please do change the `results` case as noted. Rest LGTM.

lib/TableGen/Operations.cpp
sstipano: Sorry, I should've expressed myself better. It would be correct to assume `results` is a dag, but only when there is a `results` field. In case of `OpClass` there is no `results` field. When field is not present, `getValueAsDag` returns fatal error. Seems to me things should stay as is for the `results` case.

sstipano:
I was working with a bit older llvm commit in my local checkout and I missed the switch to the new Captures attribute.
To me, it seems we should stick with what llvm exposes to intrinsics, which is currently only `captures(none)` (same as old `nocapture`). From what I see, the only other variant to be exposed is `CapturesRetOnly`. Therefore, I don't think it makes sense to have all the variations, like in memory attribute for example.

What do you think?

nhaehnle (approved):
Thanks!

PR #125, opened 2025-05-06
Update checks to use later version of Ubuntu

dstutt:


dstutt:
20.04 is no longer supported, so was causing other PRs to fail the relevant checks.

piotrAMD (approved):

PR #129, opened 2025-06-23
Fix compiler namespace warnings and deprecated getValues

Flakebi:
Fix warnings by using an anonymous namespace instead of `static` and by replacing deprecated `getValues` with `getElements`.


Flakebi:
Seems like my LLVM was too new (let’s wait a bit with this).

PR #130, opened 2025-09-16
Fix LLVM deprecation warnings

trenouf:

PR #131, opened 2025-09-25
Fix up warnings

dstutt:
Change use of ArrayRef into {}
Change deprecated getValues into getElements


dstutt:
Hmm - it seems that getValues change to getElements doesn't work here.

dstutt:
Tim already has a fix for this #130 closing

trenouf (commented):
Happy for you to go ahead with this one as I never got round to fixing up the tests in #130. LGTM, but I cannot approve.

Flakebi (approved):

PR #132, opened 2025-10-13
Support struct-backed type

piotrAMD:
Support struct-backed type, which is a custom struct type where:
1. the first field is a sentinel of integer type
2. each integer argument N passed during constructing the type is treated
   as an unsigned int encoded as a struct field:
  - intN (integer with N bit width) for positive N (up to MAX_INT_BITS)
  - empty structure for N == 0

For example
	StructBackedType::get(ctx, 1, 0, 2)
gets encoded as
	%sb_1.0.2. = type { i41, i1, {}, i2 }

Co-authored-by: Nicolai Hähnle <nicolai.haehnle@amd.com>

### DIFF
--- a/include/llvm-dialects/Dialect/Dialect.td
+++ b/include/llvm-dialects/Dialect/Dialect.td
@@ -215,6 +215,9 @@ def F64 : SpecialBuiltinType<"Double">;
 def F32 : SpecialBuiltinType<"Float">;
 def F16 : SpecialBuiltinType<"Half">;
 
+def repr_targetext;
+def repr_struct;
+
 /// All types that are defined by a dialect are derived from this class.
 class DialectType<Dialect dialect_, string mnemonic_> : Type, Predicate {
   dag typeArguments = ?;
@@ -229,6 +232,14 @@ class DialectType<Dialect dialect_, string mnemonic_> : Type, Predicate {
 
   string summary = ?;
   string description = ?;
+
+  /// How the dialect type is represented in LLVM IR:
+  ///  - (repr_targetext): use a TargetExtType (the default)
+  ///  - (repr_struct <type>): use a StructType with the given type as the
+  ///    discriminant; the discriminant should be a type that cannot naturally
+  ///    appear elsewhere, e.g. (repr_struct (IntegerType 41))
+  dag representation = (repr_targetext);
+  string structPrefix = "";
 }
 
 def and;
--- a/include/llvm-dialects/Dialect/Dialect.td
+++ b/include/llvm-dialects/Dialect/Dialect.td
@@ -239,7 +239,6 @@ class DialectType<Dialect dialect_, string mnemonic_> : Type, Predicate {
   ///    discriminant; the discriminant should be a type that cannot naturally
   ///    appear elsewhere, e.g. (repr_struct (IntegerType 41))
   dag representation = (repr_targetext);
-  string structPrefix = "";
 }
 
 def and;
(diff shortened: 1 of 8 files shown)


lib/TableGen/DialectType.cpp
```diff
@@ -44,6 +44,28 @@ bool DialectType::init(raw_ostream &errs, GenDialectsContext &context,
...
+          llvm::cast<llvm::IntInit>(
+              llvm::cast<llvm::DagInit>(dag->getArg(0))->getArg(0))
+              ->getValue();
+    }
+  }
+  if (auto *p = record->getValue("structPrefix"))
+    m_structPrefix = record->getValueAsString("structPrefix").str();
+  else
+    m_structPrefix =
+        (m_dialectRec->getValueAsString("name").str() + "." + m_mnemonic + ".");
```
nhaehnle: getValue will never return nullptr here.

More generally -- the mnemonic is what we use for the TargetExtType. For consistency, it's better to just remove struct prefix altogether and use the mnemonic as the prefix.

lib/TableGen/DialectType.cpp
```diff
@@ -195,76 +270,141 @@ void DialectType::emitDefinition(raw_ostream &out, GenDialect *dialect) const {
...
     }
 
-    FmtContextScope scope{fmt};
-    fmt.addSubst("type", argument.type->getCppType());
-    fmt.addSubst("name", convertToCamelFromSnakeCase(argument.name, true));
-    fmt.addSubst("expr", expr);
+    out << "  std::string __name; ::llvm::raw_string_ostream __os(__name);\n";
+    out << tgfmt("  __os << \"$0\";\n", &fmt, m_structPrefix);
+    for (const auto &getterArg : getterArgs)
+      out << "  __os << (uint64_t)" << getterArg.name << " << '.';\n";
```
nhaehnle: Slightly nicer IR output:
```suggestion
    for (const auto &getterArg : getterArgs)
      out << "  __os << '.' << (uint64_t)" << getterArg.name << ";\n";
```

lib/TableGen/DialectType.cpp
```diff
@@ -195,76 +270,141 @@ void DialectType::emitDefinition(raw_ostream &out, GenDialect *dialect) const {
...
+      if (auto *attr = dyn_cast<Attr>(argument.type)) {
+        out << tgfmt(attr->getCheck(), &fmt, getterArg.name) << '\n';
+      }
     }
 
-    FmtContextScope scope{fmt};
-    fmt.addSubst("type", argument.type->getCppType());
-    fmt.addSubst("name", convertToCamelFromSnakeCase(argument.name, true));
-    fmt.addSubst("expr", expr);
+    out << "  std::string __name; ::llvm::raw_string_ostream __os(__name);\n";
```
nhaehnle: Use `fmt.addSubst` to write `$name` and `$os` instead of the double underscores -- the double underscores are reserved for the C++ implementation and shouldn't be used.

include/llvm-dialects/Dialect/Dialect.td
```diff
@@ -229,6 +232,14 @@ class DialectType<Dialect dialect_, string mnemonic_> : Type, Predicate {
...
   string summary = ?;
   string description = ?;
+
+  /// How the dialect type is represented in LLVM IR:
+  ///  - (repr_targetext): use a TargetExtType (the default)
+  ///  - (repr_struct <type>): use a StructType with the given type as the
+  ///    discriminant; the discriminant should be a type that cannot naturally
+  ///    appear elsewhere, e.g. (repr_struct (IntegerType 41))
+  dag representation = (repr_targetext);
+  string structPrefix = "";
```
nhaehnle: As mentioned elsewhere: best to remove the structPrefix.

nhaehnle (commented):
Generally LGTM, but I do have some nits.

piotrAMD:
Thanks. Struct prefix and underscores removed.

nhaehnle (approved):
Thanks!

PR #133, opened 2025-11-04
Support enum class in struct-backed type

GinShio:
We may use enumeration class in struct backed type, which makes struct type more semantic. However, we need to force these values ​​to be cast from integers to the target type.

### DIFF
--- a/lib/TableGen/DialectType.cpp
+++ b/lib/TableGen/DialectType.cpp
@@ -201,17 +201,24 @@ void DialectType::emitDeclaration(raw_ostream &out, GenDialect *dialect) const {
     out << "      static bool classof(const ::llvm::Type *t);\n\n";
 
     unsigned fieldIdx = 1; // sentinel
+    auto getCastExpr = [&fmt](const NamedValue &argument,
+                              llvm::StringRef expr) -> std::string {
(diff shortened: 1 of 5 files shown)


lib/TableGen/DialectType.cpp
```diff
@@ -309,10 +309,10 @@ void DialectType::emitDefinition(raw_ostream &out, GenDialect *dialect) const {
 
     for (const auto &getterArg : getterArgs) {
       out << tgfmt(R"(
-  if ($0 == 0)
+  if ((uint64_t)$0 == 0)
     $fields.push_back(::llvm::StructType::get($_context));
   else
-    $fields.push_back(::llvm::IntegerType::get($_context, $0));
+    $fields.push_back(::llvm::IntegerType::get($_context, (uint64_t)$0));
```
Flakebi: The conversions here and above should use `getToUnsigned` and `getFromUnsigned`

GinShio: Thanks, it looks better.

nhaehnle (approved):
LGTM

GinShio:
Hi members, sorry, I cannot merge this...

PR #135, opened 2026-01-08
Add optional type name override

piotrAMD:
Add llvmTypeNameOverride property to DialectType to allow overriding the LLVM IR type name generated for dialect types. This is useful when the natural dialect+mnemonic name does not match backend conventions.


nhaehnle (approved):
LGTM. For future reference, some of these tests could have perhaps been done a little more elegantly using the ExampleDialect and textual comparison with the produced LLVM IR, but it doesn't matter much.

PR #136, opened 2026-01-27
Support structured GEP

jundaliuAMD:


jundaliuAMD:
Ping

nhaehnle (approved):

jundaliuAMD:
Need help with merging - github doesn't provide a merge button for me :(

nhaehnle:
I also missed that the workflows weren't triggered... they should run now.

nhaehnle:
Feel free to ping me again if this falls off the radar

lib/Dialect/OpDescription.cpp
```diff
@@ -244,6 +244,10 @@ HANDLE_INTRINSIC_DESC_OPCODE_SET(AnyMemCpyInst, Intrinsic::memcpy,
 HANDLE_INTRINSIC_DESC_OPCODE_SET(AnyMemMoveInst, Intrinsic::memmove,
                                  Intrinsic::memmove_element_unordered_atomic)
 
+#if !LLVM_MAIN_REVISION || LLVM_MAIN_REVISION >= 566206
```
Flakebi, line 248: Given the other code here, I think it would be good to also surround this with a `#if HAVE_LLVM_VERSION_MAJOR >= 13`

jundaliuAMD: Why check version **13**? Isn't the latest LLVM major version that contains this op **22**?

Flakebi: Whoops, good catch, I meant to write **2**3.
If I looked it up correctly, the 22 release did not contain structured geps so 23 (and current git, which defines version_major as 23) support it.

Flakebi (approved):
Thanks! (I’ll merge once CI ran through)

PR #137, opened 2026-03-10
Do not add null byte for immutable strings

jundaliuAMD:
`CreateGlobalString` will add a null byte to the end of string by default, which prevents us from getting the exact same string back as we set.

This makes it hard to use for cases where we need exact string matching, such as function names lookup.


Flakebi (approved):
Makes sense to me, thanks!

PR #138, opened 2026-03-23
Add class member function support for addSet

jiaolu:
Add class member function support


Flakebi:
Change looks good to me. Can you please add a simple use of that in the example?

Flakebi (approved):
Thanks

PR #139, opened 2026-04-01
Update tests to not use Br instruction

mariusz-sikora-at-amd:
Upstream LLVM split Br into CondBr/UncondBr
https://github.com/llvm/llvm-project/pull/184027

### DIFF
--- a/test/unit/interface/OpMapTests.cpp
+++ b/test/unit/interface/OpMapTests.cpp
@@ -34,17 +34,17 @@ TEST(OpMapBasicTestsName, CoreOpContainsTests) {
   OpMap<StringRef> map;
 
   OpDescription retDesc = OpDescription::fromCoreOp(Instruction::Ret);
-  OpDescription brDesc = OpDescription::fromCoreOp(Instruction::Br);
+  OpDescription callBrDesc = OpDescription::fromCoreOp(Instruction::CallBr);
   map[retDesc] = "RetInst";
 
   EXPECT_TRUE(map.containsCoreOp(Instruction::Ret));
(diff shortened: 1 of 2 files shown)


test/unit/interface/OpMapTests.cpp
```diff
@@ -34,17 +34,17 @@ TEST(OpMapBasicTestsName, CoreOpContainsTests) {
   OpMap<StringRef> map;
 
   OpDescription retDesc = OpDescription::fromCoreOp(Instruction::Ret);
-  OpDescription brDesc = OpDescription::fromCoreOp(Instruction::Br);
```
mariusz-sikora-at-amd, line 37: I think it is safe to change this test, the main purpose of this test is not to test `Br` instruction. `Ret` and `Br` are (were) two first Core LLVM instructions.

dstutt:
Is this safe to submit without guards for llvm main revision (I think yes?)

dstutt (approved):

mariusz-sikora-at-amd:
> Is this safe to submit without guards for llvm main revision (I think yes?)

I think it is safe. This test was just taking two first instructions from LLVM Core list of instructions. It is not testing `Br` or `Ret` in any way. You can use any other instructions. I also removed magic numbers from other test.

mariusz-sikora-at-amd:
> Is this safe to submit without guards for llvm main revision (I think yes?)

We could use llvm main revision and test `Br` and `UncondBr/CondBr` but this is not the main point of this test, so I think it would be better just to change `Br` to something different.

mariusz-sikora-at-amd:
I don't see merge button. Someone else needs to merge this PR.

PR #140, opened 2026-04-28
Add support for visiting op classes

jasilvanus:
Op classes are sets of ops, providing a common base class in the op class hierarchy,
which can be used to group sets of ops with a common property.

Before this patch, op class only support `isa<>` checks, but not visiting.

This was because visiting relies on `OpDescription::get<OpT>`, which is only supported
for concrete ops, but not op classes.

This patch changes that by introducing a new `OpDescription::getAll<OpT>` template
which returns an `ArrayRef` of OpDescriptions. The existing `get<OpT>` template
is implemented in terms of the new `getAll` template.

Visitor code is changed to use `getAll` instead, and handle non-trivial array refs
(representing op classes) accordingly, using the already existing mechanism for op sets.

I wondered whether it would be better to remove the existing `get` method and migrate everything to the new `getAll` method (and renaming it back to `get`). However, that would require to add support for op classes in all places that rely on `OpDescription`s and `get`, which is more than just visitors. Thus I chose to add a new method for this.

Also, we could instead provide a generic implementation of `get` in terms of `getAll`, because all implementations of `get` are the same: They return `getAll()[0]`. However, that would also add such methods for op classes, which we don't want. We could `assert` that the length is one, but that turns a link-time error into a debug-only assertion failure, which is strictly worse. Maybe we could add some more traits that allow to detect op classes, but I didn't want to complicate things further.

### DIFF
--- a/include/llvm-dialects/Dialect/OpDescription.h
+++ b/include/llvm-dialects/Dialect/OpDescription.h
@@ -65,7 +65,13 @@ public:
     return m_kind == Kind::Dialect || m_kind == Kind::DialectWithOverloads;
   }
 
+  // Only supported for concrete ops, not for op classes.
   template <typename OpT> static const OpDescription &get();
+
+  // For concrete ops, returns a 1-element array containing the result of get().
+  // For op classes, returns an array containing the descriptions of all
+  // concrete ops that belong to this op class.
+  template <typename OpT> static llvm::ArrayRef<OpDescription> getAll();
 
   Kind getKind() const { return m_kind; }
 
--- a/lib/Dialect/OpDescription.cpp
+++ b/lib/Dialect/OpDescription.cpp
@@ -112,6 +112,10 @@ template <> const OpDescription &OpDescription::get<UnaryInstruction>() {
   return desc;
 }
 
+template <> ArrayRef<OpDescription> OpDescription::getAll<UnaryInstruction>() {
+  return get<UnaryInstruction>();
+}
+
 template <> const OpDescription &OpDescription::get<BinaryOperator>() {
   static unsigned opcodes[] = {
 #define HANDLE_BINARY_INST(num, opcode, Class) Instruction::opcode,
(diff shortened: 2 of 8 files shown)


test/unit/interface/VisitorIRTests.cpp
```diff
@@ -0,0 +1,87 @@
+/*
+ ***********************************************************************************************************************
+ * Copyright (c) 2023 Advanced Micro Devices, Inc. All Rights Reserved.
```
Flakebi: `Copyright (c) Advanced Micro Devices, Inc., or its affiliates.` or update the year

include/llvm-dialects/Dialect/OpDescription.h
```diff
@@ -65,8 +65,14 @@ class OpDescription {
     return m_kind == Kind::Dialect || m_kind == Kind::DialectWithOverloads;
   }
 
+  // Only supported for concrete, non-op-class ops.
   template <typename OpT> static const OpDescription &get();
 
+  // For concrete, non-op-class ops, returns an array of length one that
```
Flakebi: non-op-class ops → non-class ops or “concrete ops, not op classes”?

lib/Dialect/OpDescription.cpp
```diff
@@ -108,40 +108,62 @@ template <> const OpDescription &OpDescription::get<UnaryInstruction>() {
 #define HANDLE_CAST_INST(num, opcode, Class) Instruction::opcode,
 #include "llvm/IR/Instruction.def"
   };
-  static const OpDescription desc{Kind::Core, opcodes};
+  static const std::array<OpDescription, 1> desc{
```
Flakebi: Unless needed elsewhere in that form, this can stay a `OpDescription` variable and the ArrayRef can be constructed on the fly.

Flakebi (approved):
Thanks, that’s a nice diff!

